PR #18, opened 2020-05-29
add gateway and gamemode filter to all matches

abeninski:


W3ChampionsStatisticService/Matches/MatchRepository.cs
```diff
@@ -146,6 +144,15 @@ private PlayerScore CreateDetail(PlayerBlizzard playerBlizzard)
             return events;
         }
 
+        public Task<long> Count(
+            GateWay gateWay = GateWay.Undefined,
```
gabcinder2004, line 148: I may misunderstand the full purpose, but do you think **Gateway.Any** or **GameMode.Any** would be more clear in the code than undefined?

abeninskibede:
You are totally right. I didn't see your comment. We should ask modmoto what he thinks. Maybe it's some kind of pattern?

modmoto:
yes, any is fine, i just took the undefined from the ui once, but any is
more clear in that regard.

Am Sa., 30. Mai 2020 um 07:15 Uhr schrieb Aleksandar Beninski <
notifications@github.com>:

PR #37, opened 2020-06-03
Don't ignore matchId for json output

Qeenon:
Don't ignore matchId for json output
This field is very important to be able to compare ongoing and passed matches

### DIFF
--- a/W3ChampionsStatisticService/Matches/Matchup.cs
+++ b/W3ChampionsStatisticService/Matches/Matchup.cs
@@ -16,7 +16,7 @@ namespace W3ChampionsStatisticService.Matches
 
         [JsonPropertyName("id")]
         public string ObjectId => Id.ToString();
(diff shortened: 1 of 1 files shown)


Qeenon:
generally I'd change exposed json field `id` to point MatchId instead of ids of start and end events but that change may affect something else I'm not aware about and should be discussed.

modmoto:
also the ongoing matches also use our internal id, do not know why that would help here

cnd:
@modmoto my problem is that ongoing matches has ids `Id = matchStartedEvent.Id,` and finished matches has ids `Id = matchFinishedEvent.Id,` which are different things.
I need something to find finished match if I had access to exposed ongoing one

cnd:
current change sadly indeed will not help me since API is working with `matchFinishedEvent.Id` as far as I understand and even if I will know match id it will not help me to request it...

PR #47, opened 2020-06-13
Fixed Seasons Filter Match History and Search Bug

cinderjav:


cinderjav:
Just realized I have to fix some tests

PR #70, opened 2020-09-01
Fix MMRs being counted twice

alocate:
the math.abs(x-r) < 25 part is looking at intervals of length 50


modmoto:
Did you check how that looks in the UI? It has been some time that I had a look at that, but would that not create half the mmr bars in the UI? Not so sure, that we do not have anything hard coded in the UI (e.g. the 25mmr steps).

modmoto:
thanks for the PR though =D

abeninski:
Yo Toxi adding first PR!

modmoto:
ah that is toxi =D

alocate:
now should be good without touching anything else =D?

PR #90, opened 2020-10-30
Extend mmr distribution

jensk1:
I extended the Load  mrrs function. Now its possible to have parameters of Gateway and Gamemode in the API.
So the issue in the UI can be handled https://github.com/w3champions/w3champions-ui/issues/242

The default values are Gateway Europe and Gamemode 1v1 as this seemed to be (or should be) the default. So it should be backward compatible with any functions in UI at the moment.

All tests in my projects passed, but it seems LoadMMRsDitribution function is not involved in any test.
Also tested on local Mongodb and works.

If you want to have any further changes, let me know pls.

When there are cases the old function is still needed I could change it to an overload function.

### DIFF
--- a/W3ChampionsStatisticService/PlayerProfiles/PlayerRepository.cs
+++ b/W3ChampionsStatisticService/PlayerProfiles/PlayerRepository.cs
@@ -58,11 +58,14 @@ namespace W3ChampionsStatisticService.PlayerProfiles
             return UpsertMany(winrate);
         }
 
-        public async Task<List<int>> LoadMmrs(int season)
+
+        public async Task<List<int>> LoadMmrs(int season, GateWay gateWay, GameMode gameMode)
         {
             var mongoCollection = CreateCollection<PlayerOverview>();
             var mmrs = await mongoCollection
-                .Find(p => p.Season == season)
+                .Find(p => p.Season == season &&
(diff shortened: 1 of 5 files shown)


modmoto:
seems fine to me, but make sure, that the ui releases in parallel, because it will break the view and only show season 1/europe. A added tests would be nice aswell =)

alocate:
that's really awesome news =)

jensk1:
I wrote simple test cases looping over the enums, checking if every combination returns something. Seems like the Testbase creating new Database where no sample data can be fetched. Do I have to insert the data i want to check on first? What would be the quickest way

jensk1:
Okay, I just figured out I was in the wrong Collection. And the answer is yes, for testing you have to insert your own dummy data to test on

jensk1:
There is now a Testcase which passes and iterates over the possible commbination of Enums.
I leaved out 2v2 events cause of the special case that there hase to be 4 players: https://github.com/w3champions/w3champions-statistic-service/blob/master/W3ChampionsStatisticService/Ladder/PlayOverviewHandler.cs#L33

I will check for the UI changes next time :)

modmoto:
nice, thanks a lot =) i will check the ui pr now aswell. Backend should be on test in a few minutes.

PR #91, opened 2020-11-03
search profile if no ranking is found

knumf:


knumf:
connected to https://github.com/w3champions/w3champions-ui/pull/269

modmoto:
I dont really get what that would do. If I search for "trollolol" I get an response that contains this name but no other information? When I click this thing, I would be redirected to where? Profile Page will not load for sure :/

I think what you are trying to do is to make people that never played visible, right? I think the better way to do this would be to create a new profile, when someone logs in and never played a game yet. Because this means, that the battletag exists and the player should have a 0:0 Account. but then he could also join Clans already and use the rest of the page.

That would be cleaner and would partially fix the solution for searching players that are unranked/have not played a game. Also finding players that are unranked would be nice. This could be implemented without the account creation aswell.

modmoto:
Sorry, i just realized, that I already implemented the "create profile on first login" part. So we would just have to search in other collections. I think the `Task<List<PlayerOverallStats>> SearchForPlayer(string search);` in the player repo could be used for that. because we generate a `PlayerOverallStats` Readmodel when the user logs in for the first time (see `RegistrationHandler`)

knumf:
hi, thanks for your input! =) hmm you do see that the collection and method you refer to is the one being used if rank is not found? i chose to squeeze the result into the existing rank model to avoid creating a new model / contract for the ui component to deal with. maybe the cleaner solution would be to create a common search response model in this controller action instead of returning the rank model in the first place?

i'm not sure about your trollol example, it would only return a response if the PlayerOverallStats is found. And the battletag would then be included and used to route to profile. 

profileexist
notfound

abeninskibede:
Issue: https://github.com/w3champions/w3champions-ui/issues/203

abeninski:
Issue: https://github.com/w3champions/w3champions-ui/issues/203

modmoto:
> hmm you do see that the collection and method you refer to is the one being used if rank is not found?

ah jeah, now I do xD. Yeah, then this seems fine. I somehow was looking at it the wrong way.

PR #92, opened 2020-11-06
MMR Timeline for individual players

weidenbach:
This relates to issue #215 in UI (https://github.com/w3champions/w3champions-ui/issues/215)

The goal is to create an interface for the frontend to load MMR data to create a graph. With the parameters battletag, race, gateway and season, specific data could be easily queried from the backend to display graphs for various constallations.

As I'm new to the project, I'm not sure whether the mmr values used are the correct ones. In the "PlayerMmrTimelineHandler", once im using "Match" (Update()) and once "Matchup" (CreateMmrTimeline()) to get the mmr values. Not sure if this is problematic. From "Match" I can get a "PlayerMMrChange", which has both "mmr" and "updatedMmr". Which one to use?

I added a test, but I'm getting weird values for the mmr numbers. They seem to be random, not sure why.

Feedback would be greatly appreciated!

### DIFF
--- a/W3ChampionsStatisticService/PlayerProfiles/MmrRankingStats/PlayerMmrTimeline.cs
+++ b/W3ChampionsStatisticService/PlayerProfiles/MmrRankingStats/PlayerMmrTimeline.cs
@@ -1,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+using W3ChampionsStatisticService.CommonValueObjects;
+using System.Linq;
+using System.Threading.Tasks;
+using W3ChampionsStatisticService.ReadModelBase;
+
+namespace W3ChampionsStatisticService.PlayerProfiles.MmrRankingStats
+{
+    public class PlayerMmrTimeline : IIdentifiable
+    {
+        public PlayerMmrTimeline(string battleTag, Race race, GateWay gateWay, int season)
+        {
+            Id = $"{season}_{battleTag}_@{gateWay}_{race}";
+            BattleTag = battleTag;
+            Race = Race;
+            GateWay = gateWay;
+            Season = season;
+        }
+        public List<MmrAtTime> MmrAtTimes = new List<MmrAtTime>();
+        public string BattleTag { get; set; }
+        public Race Race { get; set; }
+        public GateWay GateWay { get; set; }
+        public int Season { get; set; }
+        public string Id { get; set; }
+
+    }
+
+    public class MmrAtTime
+    {
+        public MmrAtTime(int mmr, DateTimeOffset mmrTime) {
+            Mmr = mmr;
+            MmrTime = mmrTime;
+        }
+        public int Mmr;
+        public DateTimeOffset MmrTime;
+    }
+}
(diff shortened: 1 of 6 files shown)


alocate:
hey there,

in the mongodb I use, MMR is the MMR before a game, updated MMR is the MMR after a match. I assume it's the same here.

thanks for your time!

modmoto:
nice one, lets see how long this takes to build on test and then release it to prod. So the model can be built before we release the ui feature.

weidenbach:
Thanks guys! I wasn't expecting a direct merge, as I'm still confused about the test. Let me elaborate.

I set up three 1v1 games. All between peter and wolf.
Peter wins game 1 and 3, wolf game 2.
Hence, peter should have higher mmr after game 1 and 3 than after game 2. 
The opposite for wolf.
But the mmrs seem random. 
I added a commit that adds these mmr comparisons as asserts. The test fails mostly, sometimes passes.

My guess is that the mmr of each matchup is generated randomly and not based on the previous matchup, at least for this test I wrote. If that's the case, then the MmrTimeline works correct. 
Would be great if someone could shed some light on this behavior 😄

modmoto:
I think the matchevent that you use uses the framework to generate dtos with random data. For that you would have to set it to a fixed value. That could be the case. Also on test the handler never starts, maybe you forgot a dependcy injection?

modmoto:
do you have the test db connectionstring @weidenbach ? you can also use the db from the readme to test it locally (test and this db are the same)

weidenbach:
> 
> 
> I think the matchevent that you use uses the framework to generate dtos with random data. For that you would have to set it to a fixed value. That could be the case. Also on test the handler never starts, maybe you forgot a dependcy injection?

You were right, the generated dtos uses random data. I updated the test to use fixed mmr values. Works perfectly now. 
For me, when I run the test, the handler works fine. I inject the player- and matchrepository so the handler has the things it needs. Or do you mean something else?

I also added a sorting to the PlayerMmrTimelineHandler so the values are already presorted. So I have 2 new commits, should I add another PR for that? I'm new to opensource work so bear with me 😉

weidenbach:
> 
> 
> do you have the test db connectionstring @weidenbach ? you can also use the db from the readme to test it locally (test and this db are the same)

Hmmm you mean mongodb://176.28.16.249:3513 ? Yes, it's the default if I'm not mistaken. 
I never used mongodb or anything related, so I don't know how to test on the database. I mean, even if I am connected I would need the MatchFinishedEvent to test the handler, right? How can I use the db to test my handler?

weidenbach:
A small ping for @modmoto , in case that you missed my reply

modmoto:
sorry, was busy these days, good ping ;)

as the handler does not start, when you run it locally (look in startup.cs), you have to enable your new handler. Just copy the DI outside the if and start it like that (but of course do not check it in like that, so other people do not accidentally start syncers when they check out the project). If you are unsure, you can also clone the db locally, but as this is a new readmodel, just running it and deleting the collection + HandlerVersion afterwards should be fine. Take a look at the read model handler section in the readme to get more info on how to debug. Right now the Handler is on verison 0, or does not exists at all tbh. Which means it most likely gets an exception on the first run.

PR #95, opened 2020-11-22
Update2: MMR Timeline for individual players

weidenbach:
Adds RP functionality to the timeline.
Changes timeline so only the latest MMR/RP pair per day is in it
Refactored timeline to include RP in name

**Requires rebuilding of handler**


modmoto:
nice one, thanks for that =)

PR #99, opened 2020-11-25
Should now show all Ongoing matches in-memory

mourasman:
There was a NullPointerException happening on LoadOnGoingMatchForPlayer function. It only worked the first time because the first time it's executed the list is empty so the .Find method will return null, by default. All subsequent 2-team matches that get added, will throw an NPE because team3 and team4 are null. Added a null check to prevent this.


modmoto:
i just saw that this was merged already? I actually did the same thing right now, but with a cache, so we do not loose the data when we deploy. Should put it down to one request per minute.

mourasman:
yeah, no worries. this version is bugged and yours seems a more robust solution anyway. closing this PR. please revert my PR from the main repo.

modmoto:
i did not test it yet, though, just opened the pr so it is more visible (not just the branch)

but thanks for the PR anyway!

PR #103, opened 2020-12-02
add twitch to players in ongoing matches

SergeyKhval:
This is a step needed to implement active streams on home page and may be on ongoing matches page
image

### DIFF
--- a/W3ChampionsStatisticService/Matches/MatchQueryHandler.cs
+++ b/W3ChampionsStatisticService/Matches/MatchQueryHandler.cs
@@ -1,0 +1,39 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using W3ChampionsStatisticService.Ports;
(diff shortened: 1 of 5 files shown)


modmoto:
This would be needed to show the twitch channel on the ongoing matches? Or do we need that for something else? The matches are now in memory, because they had performance issues.  The ongoing matches are about 1/6ths of our requests. Just make sure, that we do not get any problems when loading all the user settings everytime we go on the ongoing matches page.

Cepheid-UK:
This supports https://github.com/w3champions/w3champions-ui/issues/279

SergeyKhval:
I added UI part for this. I am going to deploy it to test and take a look at performance. Is it possible to use prod DB on test to check with realtime data?

modmoto:
Yes we could, but I do not want to mess with the data on prod. We can just deploy and see what happens ;)

PR #107, opened 2020-12-08
Add random race to matchup players.

aaronjewell:
Use the player's hero selection to determine, if possible,
which race they were assigned from the random selection.
Assigned to a new member property, so that it can be used
separately for statistical computations/display. If the
player's race is not random, then null is assigned to the
random race member. If it cannot be determined which race
was assigned, then random is set for both the player race
as well as the random race.

First step required for handling w3champions/w3champions-ui#281
sanely in multiple UI components. Also, computed in the record so
that statistics can be computed at a later time.

Does not backfill previous matches.

### DIFF
--- a/W3ChampionsStatisticService/Matches/Matchup.cs
+++ b/W3ChampionsStatisticService/Matches/Matchup.cs
@@ -68,16 +68,30 @@ namespace W3ChampionsStatisticService.Matches
                 .ThenBy(x => x.team)
                 .ToList();
 
-            foreach (var resultPlayer in matchFinishedEvent.result.players)
+            if (matchFinishedEvent.result == null)
             {
-                var matchPlayer = players.First(p => p.battleTag == resultPlayer.battleTag); 
-
-                // If the player chose random for the match,
-                // set their actual randomized race from the result.
-                matchPlayer.rndRace = matchPlayer.race == Race.RnD
-                    ? matchPlayer.race.FromRaceId((RaceId)resultPlayer.raceId)
-                    : null;
+                foreach (var matchPlayer in players)
+                {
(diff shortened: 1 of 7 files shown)


aaronjewell:
C# is not my primary language, and I don't have ReSharper or similar helpers installed, so it's likely there is a more succinct way to write some of the race determination code. I would love any feedback! 😃

abeninski:
Hey! Thanks for the PR!

Heroes are not the most reliable way to determine race.
We discovered that the game reports the race, so you can use that.

In the `MatchFinishedEvent` you can see there is a `result` property and in in `raceId`
raceId: 1 = human
raceId: 2 = orc
raceId: 3 = undead
raceId: 4 = elf

aaronjewell:
Even better! Updated to use this result property.

W3ChampionsStatisticService/Matches/Matchup.cs
```diff
@@ -68,6 +68,17 @@ public static Matchup Create(MatchFinishedEvent matchFinishedEvent)
                 .ThenBy(x => x.team)
                 .ToList();
 
+            foreach (var resultPlayer in matchFinishedEvent.result.players)
```
abeninski, line 71: In some cases result is null. Can you just put 1 if and it will be fine :)

aaronjewell:
@abeninski - Thanks for catching that! How do you feel about this update? Rather than default them to a specific race for the rndRace member property when there is no result in the event, we can set it as random. That way if we want to use this for stat calculation later, it doesn't skew the results towards a specific race.

aaronjewell:
Added a test for the null result, and moved the Race enum extensions into its own file to follow the conventions of the other enums.

PR #108, opened 2020-12-10
Add map filter to ongoing and previous matches.

aaronjewell:
Defaults to the "Overall" string, which matches the accompanying
change to the front-end. A future enhancement to this feature
would be an endpoint that could retrieve the map choices for a
given Gateway/GameMode combination. The front end could then
dynamically draw the map select from this data. This would
be a better user experience, since it would reduce the number
of empty result outcomes.

Backend implementation to address part of w3champions/w3champions-ui#277


Cepheid-UK:
@modmoto can you take a look at this for https://github.com/w3champions/w3champions-ui/pull/309, if you're ok to push this to test, I can check the frontend.

modmoto:
ok i released to prod, seems good =D thanks for the PR!

PR #109, opened 2020-12-10
Add handling for missing leagues in season 0.

aaronjewell:
There are some ranks in season 0, that do not have a league
for the corresponding league constellation, e.g. 0_Europe_GM_1v1.
Rather than rely on the try/catch to catch the Linq exception on
the Single call, use SingleOrDefault specifically for season 0,
so that we avoid the bloating of the logs, but can still catch
exceptions on later seasons.

should resolve w3champions/w3champions-ui#287


abeninski:
Thank you for the PR!

PR #120, opened 2021-01-15
Remove akas endpoint - pending discussion

Cepheid-UK:


Cepheid-UK:
After discussion, this is no longer necessary.

PR #124, opened 2021-01-28
[WIP]: add ability to upvote players

SergeyKhval:
This PR adds ability to upvote a player. This will be used from Twitch extension, so I'll need to validate their JWT before allowing to upvote

### DIFF
--- a/W3ChampionsStatisticService/PersonalSettings/PersonalSettingsController.cs
+++ b/W3ChampionsStatisticService/PersonalSettings/PersonalSettingsController.cs
@@ -104,5 +104,13 @@ namespace W3ChampionsStatisticService.PersonalSettings
 
             return Ok();
         }
+
+        [HttpPost("{battleTag}/upvote")]
+        public async Task<int> UpvotePlayer(string battleTag)
+        {
+            var result = await _personalSettingsRepository.Upvote(battleTag);
+
+            return result.Upvotes;
+        }
     }
 }
--- a/W3ChampionsStatisticService/PersonalSettings/PersonalSetting.cs
+++ b/W3ChampionsStatisticService/PersonalSettings/PersonalSetting.cs
@@ -42,6 +42,8 @@ namespace W3ChampionsStatisticService.PersonalSettings
         public string CountryCode { get; set; }
 
         public string Location { get; set; }
+        
+        public int Upvotes { get; set; }
 
         public string HomePage { get; set; }
         public ProfilePicture ProfilePicture { get; set; } = ProfilePicture.Default();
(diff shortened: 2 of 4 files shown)


modmoto:
But now we can upvote mutliple times, right? Would that not be easily exploitable? The identity service hands out jwts now. So if you want to validate the person upvoting, this should be doable. I am moving the backend to use the jwts soon, so maybe we reuse that there.

W3ChampionsStatisticService/PersonalSettings/PersonalSettingsController.cs
```diff
@@ -104,5 +104,13 @@ public async Task<IActionResult> MigrateCountry()
 
             return Ok();
         }
+
+        [HttpPost("{battleTag}/upvote")]
+        public async Task<int> UpvotePlayer(string battleTag)
+        {
+            var result = await _personalSettingsRepository.Upvote(battleTag);
```
modmoto, line 111: like mentioned, multiple upvotes should not be possible

SergeyKhval:
Yes, users should be able to upvote multiple times, because I'm going to integrate this feature with Twitch bits. So they will basically spend bits to buy upvotes.

W3ChampionsStatisticService/PersonalSettings/PersonalSettingsController.cs
SergeyKhval: It should be possible. The flow is as follows:

1. User clicks on upvote button in Twitch Ext
2. They receive a confirmation from twitch that they enough bits / purchase bits
3. If the purchase is completed we send a request to our backend to upvote

modmoto: But if I am smart, I just use this endpoint independantly from twitch and give me a million upvotes? Can we add a secret that is needed for the twitch extension? Or are the extension secrets open like on a website? Or maybe do this via cors, so only twitch can acces that endpoint. Dont know. But we would need some sort of validation within that handler.

SergeyKhval: That's exactly the idea. Each request from Twitch extension will have a signed JWT. On the backend I will store Twitch secret to verify this JWT. Once verified, I can proceed with upvote

modmoto: Ok perfect, but then we wait until the jwt stuff is also included? I dont think people will really go through the code and search for this endpoint, but as it is connected to money, we should wait until the endpoint is secured. Or featuretoggle it.

SergeyKhval: Yes, no need to merge now. I'll mark this as a WIP

PR #125, opened 2021-01-30
Merge season ranked and all season unranked in ladder search.

aaronjewell:
When searching for players in the ladder, merge any players
without a ranking from all seasons with ranked players. This
allows for players from previous seasons to still have their
profiles found in search.


aaronjewell:
https://statistic-service-allow-unranked-results-from-any-season.pr.w3champions.com/ should be the PR preview url.

W3ChampionsStatisticService/Ladder/LadderController.cs
```diff
@@ -33,13 +33,17 @@ public class LadderController : ControllerBase
         public async Task<IActionResult> SearchPlayer(string searchFor, int season, GateWay gateWay = GateWay.Europe, GameMode
         gameMode = GameMode.GM_1v1)
         {
-            var playerRanks = await _rankRepository.SearchPlayerOfLeague(searchFor, season, gateWay, gameMode);
+            System.Collections.Generic.List<Rank> playerRanks;
```
modmoto, line 36: i would just make a normal import

modmoto (commented):
I dont really remember the "unrankedResponse", are we sure we dont list players twice here? If you played s1, skipped s2 and s3 and you search in s4, will i get two unranked responses here? I guess not, just wanted to make sure.

W3ChampionsStatisticService/Ladder/LadderController.cs
aaronjewell: Ah yea, that was a leftover from when I had a conditional assignment. I didnt switch it back.

PR #128, opened 2021-02-03
Back End Portion Of Adding Achievements

gDollaZ:
This was created with the intent to easily update and maintain W3Champions achievements. There were only two examples given, so the code currently works with those and some similar ones. I hope to update the code further so the engine will be able to accept more formatted achievements as specified by the team. I envision the W3Champions team being able to simply add or remove JSON objects from the Achievements.json file as they see fit, this can be accomplished quickly and easily.

This is intended to integrate with this once it is completed:
https://github.com/w3champions/website/issues/213


modmoto:
is this already mergable? Do we have dependencies to the backend?

gDollaZ:
Hello @modmoto. It should be mergable, is there an issue with it? All looked good when testing locally. I have not yet created the frontend piece for this, but intend to do so soon. I need to discuss with kenshin/riptide as I had only mentioned it to them. I want to get a solid understanding of how they would like this data to be viewed.

modmoto:
I dont know if we can generalize this or if we overcomplicate this by trying to make it generic. As this is tied to a release anyway, I think we can also just code the rules as classes from an interface `IRule` or something like that. So in that case, I would just have two classes, `MapWith25Wins` and `Played ATWith5DifferentPartners`. You can still make it easy to implement, with a base class `IWinsNeededRule` like that:
```
public interface IWinsNeededRule 
{
   int WinsNeeded { get; }
   Race Race { get; }
}
```

Expanding upon that, I think we should make the Achievements a Readmodel like the other Readmodels and then go over all the events to calculate them only once. Like loading all matches of a player everytime you open a profile might cause performance issues. And of course it deviates from the usual way of how we are doing it right now.

So in the Readmodelway, we would have one handler for each rule, so we can extend rules. I would still save the Achievements in one collection, but that could just be class like this:
```pulic class PlayerAchievements
{ 
   public List<string> CompletedAchievements { get; set; }
}
```
Make sure to make the save atomic, as we would be writing this from mutliple threads. Then you just load this on the route and off we go.

Is it clear how I would imagine this?

gDollaZ:
I will have to agree with you on all of this. I was not sure if anyone wanted these values to be saved in the DB, so that is why I went about it the way I did. Thanks for the feedback, I will implement these changes as you suggested.

Cepheid-UK:
This seems very cool.

The way you've done rules is by an array of strings, and you are splitting the strings themselves?

I feel like perhaps there's a better way to handle achievement rules? Maybe a "Rule" type that we can inherit, it might make it easier for testing?

Something feels a bit urgh with storing our rules logic in strings like that.

I don't know, just spitballing ideas here.

PR #129, opened 2021-02-11
Add country, country code, and location to matchups on event handle.

aaronjewell:
In some cases, players only have a non-null Location field in their personal settings. So when building the matchup records from their events, update the Location field for all players.

This will work for new matchup's going forward, but does not handle past Matchup's.

Resolves w3champions/website#264 ... for future matches only.

### DIFF
--- a/W3ChampionsStatisticService/Matches/MatchReadModelHandler.cs
+++ b/W3ChampionsStatisticService/Matches/MatchReadModelHandler.cs
@@ -8,12 +8,15 @@ namespace W3ChampionsStatisticService.Matches
     public class MatchReadModelHandler : IReadModelHandler
     {
         private readonly IMatchRepository _matchRepository;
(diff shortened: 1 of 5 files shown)


aaronjewell:
Made a change to get the location data if it doesn't exist at controller run time. Additionally it will update the matchup record with the location data to prevent the need for a lookup on later loads.

The side effect here is the first views per page will be **significantly** slowed down for the first load. I considered adding a "fire and forget" style method call here, but with context and scopes, I don't think the trade-off in complexity is worth it.

I could be convinced that we don't update the matchup's here. On my local testing with the test database, without doing the updates, viewing a page of matches increased from 1.7s to 2.0s by adding the location lookup. Over time, the additional load on the system would greatly outweigh the single load cost of updates.

This approach also seemed simpler instead of attempting to reconcile the data in the document store.

PR #144, opened 2021-06-27
WIP: Add achievements

gDollaZ:
I redid this using the readmodel way, as suggested by @modmoto .

Marked as WIP for 2 reasons:
Would like to test this upon a real MatchFinishedEvent (is that possible?)
Should I add some more achievements, or can we use these 2 as a starting point and then add more later?

If you would like anything refactored or have any criticisms please let me know and I will fix it up.


W3ChampionsStatisticService/Achievements/AchievementRepositoryHandler.cs
```diff
@@ -0,0 +1,272 @@
...
+            string battleTag,
+            List<Matchup> matches,
+            bool isFirstRun
+            ){
+            if (matches == null){matches = await GetAllPlayerMatches(playerOverallStats);}
+            if (battleTag == null){battleTag = playerOverallStats.BattleTag;}
+            if (isFirstRun){playerAchievement.Counter = new Dictionary<string, int>();}
+        
+            var achievementProgressCounter = playerAchievement.Counter;
+            switch(playerAchievement.Id){
```
modmoto, line 174: i would put this logic in the achievement classes and make it an interface or abstract class. then you would only have to call a method like "IsAchieved(player, game)" or something like that (depending in what you need). Like that you could move the logic in the achievements and when we want to implement new achievements, we could just implement a class and be done with it. Issue also will be (right now): what happens if I add a new achievement? I would have to whipe and recalculate all achievements, right?

W3ChampionsStatisticService/Achievements/AchievementRepositoryHandler.cs
```diff
@@ -0,0 +1,272 @@
...
+        private List<Achievement> UpdateCurrentPlayerAchievementList(List<Achievement> currentAchievementsList){
+            if (currentAchievementsList == null) {currentAchievementsList = new List<Achievement>();}
+                var currentListIds = new List<long>();
+                foreach(Achievement achievement in currentAchievementsList){
+                    currentListIds.Add(achievement.Id);
+                }
+                foreach(long activeAchievementsId in ActiveAchievementIds){
+                    if(!currentListIds.Contains(activeAchievementsId)){
+                        switch (activeAchievementsId) {
+                            case 0:
```
modmoto, line 131: if you have an abstract class/interface, those switch cases can go. This tends to explode heavily, would be 100 cases for 100 achievements.

modmoto (commented):
I think we should also check what achievements we want to implement. Two achievements are not that cool ;) Maybe ask the other what ideas would be good? Maybe on played heroes or something like that? We have some data for that.

W3ChampionsStatisticService/Achievements/AchievementRepositoryHandler.cs
gDollaZ: You wouldn't need to recalculate all achievements. This will add the new one BUT we would need to grab all the player information again to run through it and look for the data we need.

Cepheid-UK:
@gDollaZ are you still dev'ing this on this branch?

PR #152, opened 2021-11-15
Add put/delete portraits endpoint for SpecialPictures

Cepheid-UK:
I've added put/delete endpoints that *should* add or delete portraits from the `SpecialPictures` field of `PersonalSettings` collection.

I've made this part of the `PersonalSettingsCommandHandler` class as this seemed relevant domain.

I also removed some code entropy.

Would appreciate a check as this as I'm still not confident with upserting.

I may add some integration tests tomorrow.


W3ChampionsStatisticService/PersonalSettings/PersonalSettingsCommandHandler.cs
```diff
@@ -35,19 +36,35 @@ public async Task<bool> UpdatePicture(string battleTag, SetPictureCommand comman
...
+            var settings = await _personalSettingsRepository.LoadMany(command.BnetTags.ToArray());
+            
+            foreach (var playerSettings in settings)
             {
-                var playerProfile = await _playerRepository.LoadPlayerProfile(battleTag);
-                setting = new PersonalSetting(battleTag, new List<PlayerOverallStats> { playerProfile });
+                var specialPortraitsList = new List<SpecialPicture>(playerSettings.SpecialPictures);
+                foreach (var portrait in command.Portraits)
+                {
+                    specialPortraitsList.Add(new SpecialPicture(portrait, command.Tooltip));
```
abeninski, line 48: You probably want to check if portrait does not already exists. If it does, should not be added twice

PR #160, opened 2022-01-15
Added support for player race filter in match history

Raswanson01:


Raswanson01:
Not sure why it's showing merge conflicts, I just added a param like I did for the opponent race filter, but both changes should be accepted on this one

PR #180, opened 2022-07-14
[W3C-108] Add timeframe filter for player matches search

davidkan2003:
Added ability to filter by timeframe when searching player matches


Cepheid-UK:
If you can either post some proof of manually testing this, or an integration test that passes (preferred) then I can merge!

davidkan2003:
> If you can either post some proof of manually testing this, or an integration test that passes (preferred) then I can merge!

image
Here is an image with the results of local testing. You can see that specifying a timespan still reeturns a response with less matches. I have checked the results and they are in compliance with the timespan specified. Sadly, I am very vaguely familiar with unit tests, so I can't give you anything better.

PR #181, opened 2022-07-15
Fix map names in maps statistics

davidkan2003:


abeninski:
This is the obvious fix but we need to decide what to do with existing data

Cepheid-UK:
should we merge this then merge the data for this season?

PR #199, opened 2022-08-30
Caching functionality for count and matchup

ShaDeFaDe:
Added caching for heavy load endpoints. First 500 records for all gamemodes are cached. Count is also cached.


gustav87:
Closing due to inactivity.

PR #200, opened 2022-09-01
[W3C-1] Player global search

PlayLikeNeverB4:
Jira issue: https://w3champions.atlassian.net/browse/W3C-1
Github issue: https://github.com/w3champions/website/issues/493

Frontend PR: https://github.com/w3champions/website/pull/536

**Changes**

Added a new endpoint `/api/players/global-search` with pagination.

Screenshot 2022-09-01 at 21 00 43

It looks in the `PersonalSetting` collection and it joins with `PlayerOverallStats` to get the season participations.

Not sure how to evaluate the performance (never used MongoDB before), but with a small page size it can't be too bad.

I set the page size to 20 thinking the searched player will most likely be among the first entries, so no need to load many of them.

Added a test for this feature.

**Related bug**

There is a bug that is highlighted by this feature: when accessing some profiles (for example `Grubby#1232`), it tries to find the `PlayerOverallStats`, but some player don't have it. This bug is on prod as well, but you can't easily get to that profile.

https://github.com/w3champions/website-backend/blob/38875bf30f26a302d8b391088851863599f63b99/W3ChampionsStatisticService/PlayerProfiles/PlayersController.cs#L47


Cepheid-UK:
Fantastic work, looks really great.

One thought - We currently cache PersonalSettings via this service:

https://github.com/w3champions/website-backend/blob/01cec7fcf9aab364fdf7bad6707076e37c24a178/W3ChampionsStatisticService/Services/PersonalSettingsProvider.cs

Is there any way you can refactor to use that cache, and do the sorting with LINQ, saving us on db requests?

PlayLikeNeverB4:
@Cepheid-UK I fixed the feedback, you can review.

> Is there any way you can refactor to use that cache

Just so I understand... So we load the entire list of ~35k `PersonalSetting` objects in memory and then for each API request we filter the records we want. Is this useful because the db is the bottleneck and we can spare the extra CPU?

Cepheid-UK:
I had totally forgotten about this, we good to merge?

PlayLikeNeverB4:
@Cepheid-UK From my point of view, yes.

PR #215, opened 2022-11-29
Only fetch last 2 weeks of finished matches

gustav87:


Cepheid-UK:
looks like other changes have superceded the need for this.

PR #225, opened 2023-02-16
Deploy map descriptions on statistics

Uladzim1r:
- Remove static cache
- Add cache configuration
- Add new model with map name
- Added service to update map name according to current season map names
- Remove some of async locks


W3ChampionsStatisticService/Services/PlayerAkaProvider.cs
```diff
@@ -11,40 +11,37 @@ namespace W3ChampionsStatisticService.Services
...
-            catch
-            {
-                return new List<PlayerAka>();
-            }
+            _userAccountsCache = userAccountsCache;
         }
 
-        public static async Task<List<PlayerAka>> FetchAkas() // list of all Akas requested from W3info
-        { 
+        public static async Task<List<PlayerAka>> FetchBattleNetAccounts() // list of all Akas requested from W3info
```
Cepheid-UK, line 21: This function isn't getting bnet accounts, it's getting a human-readable reference for us to consult bnet accounts against.

W3ChampionsStatisticService/Services/PlayerAkaProvider.cs
```diff
@@ -53,10 +50,10 @@ public static List<PlayerAka> FetchAkasSync()
             return new Player(); // returns an default values if they are not in the database
         }
 
-        public Player GetAkaDataByPreferences(string battletag, PersonalSetting settings)
+        public async Task<Player> GetAkaDataByPreferences(string battletag, PersonalSetting settings)
```
Cepheid-UK, line 53: We might as well add an "Async" on the end of this function too

W3ChampionsStatisticService/PlayerProfiles/PlayersController.cs
```diff
@@ -61,16 +61,16 @@ public async Task<IActionResult> GetPlayer([FromRoute] string battleTag, [FromQu
...
-                
+
                 player = PlayerOverallStats.Create(battleTag);
             }
-            
+
             // Akas are stored in cache - preferences for showing akas are stored in DB
             var settings = await _personalSettingsRepository.Load(battleTag);
-            player.PlayerAkaData = _playerAkaProvider.GetAkaDataByPreferences(battleTag, settings);
+            player.PlayerAkaData = await _playerAkaProvider.GetAkaDataByPreferences(battleTag, settings);
```
Cepheid-UK, line 70: probably we should rename this `GetAkaDataByPreferencesAsync` if we're doing that elsewhere

W3ChampionsStatisticService/Startup.cs
```diff
@@ -76,9 +79,23 @@ public void ConfigureServices(IServiceCollection services)
...
             services.AddSingleton<TrackingService>();
-            services.AddSingleton<PlayerAkaProvider>();
-            services.AddSingleton<PersonalSettingsProvider>();
-            services.AddSingleton<MatchmakingProvider>();
+            services.AddTransient<PlayerAkaProvider>();
+            services.AddTransient<PersonalSettingsProvider>();
+            services.AddTransient<MatchmakingProvider>();
+
+            services.AddMemoryCache();
+            services.AddTransient(typeof(ICacheData<>), typeof(InMemoryCacheData<>));
```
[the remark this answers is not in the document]
Uladzim1r: Most developers make a wrong decision about services lifetime. 
Almost all services, controllers we have in a single representation, but everything should be Transient by default to support correct dependency injection. 
In fact singletons can only be cache, http clients, maybe db connection. Because they bound to server instance
Scoped only User Information, Or request cache between services.
Our service should be Transient.
services.AddMemoryCache(); 
this line add .NET memory cache as singleton, but it is hidden behind intentionally.

W3ChampionsStatisticService/Startup.cs
```diff
@@ -61,7 +64,7 @@ public void ConfigureServices(IServiceCollection services)
             });
 
             var startHandlers = Environment.GetEnvironmentVariable("START_HANDLERS");
-            var mongoConnectionString = Environment.GetEnvironmentVariable("MONGO_CONNECTION_STRING")  ?? "mongodb://157.90.1.251:3513"; // "mongodb://localhost:27017";
+            var mongoConnectionString = Environment.GetEnvironmentVariable("MONGO_CONNECTION_STRING") ?? "mongodb://localhost:27016";
```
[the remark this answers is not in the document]
Uladzim1r: Committed by mistake

W3ChampionsStatisticService/PlayerProfiles/PlayersController.cs
Uladzim1r: Done

W3ChampionsStatisticService/Services/PlayerAkaProvider.cs
Uladzim1r: Renamed, tried to make it readable but didn't understand currently)

Uladzim1r:
Renamed to ICachedDataProvider to make it explicit

PR #266, opened 2023-08-29
remove try/catch from W3CAuthenticationService and add it on the call…

gustav87:
…ing code. Do not validate token lifetime except for admins

### DIFF
--- a/W3ChampionsStatisticService/WebApi/ActionFilters/CheckIfBattleTagBelongsToAuthCodeFilter.cs
+++ b/W3ChampionsStatisticService/WebApi/ActionFilters/CheckIfBattleTagBelongsToAuthCodeFilter.cs
@@ -1,8 +1,10 @@
-﻿using System.Linq;
+﻿using System;
+using System.Linq;
 using System.Threading.Tasks;
 using System.Web;
 using Microsoft.AspNetCore.Mvc;
 using Microsoft.AspNetCore.Mvc.Filters;
+using Microsoft.IdentityModel.Tokens;
 using W3ChampionsStatisticService.WebApi.ExceptionFilters;
 
 namespace W3ChampionsStatisticService.WebApi.ActionFilters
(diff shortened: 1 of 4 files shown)


W3ChampionsStatisticService/WebApi/ActionFilters/CheckIfBattleTagBelongsToAuthCodeFilter.cs
```diff
@@ -22,21 +24,25 @@ public async Task OnActionExecutionAsync(ActionExecutingContext context, ActionE
...
-                if (
-                    res != null
-                    && !string.IsNullOrEmpty(btagString)
-                    && btagString.Equals(res.BattleTag))
-                {
-                    context.ActionArguments["battleTag"] = res.BattleTag;
-                    await next.Invoke();
+                    if (!string.IsNullOrEmpty(btagString) && btagString.Equals(res.BattleTag))
+                    {
+                        context.ActionArguments["battleTag"] = res.BattleTag;
```
Cepheid-UK, line 33: Need to handle an exception if res is null, it will return 401 and it won't be clear what happened.
If Res = null, we should return 502, which will indicate the issue is in ID service.

W3ChampionsStatisticService/WebApi/ActionFilters/CheckIfBattleTagBelongsToAuthCodeFilter.cs
```diff
@@ -22,21 +24,25 @@ public async Task OnActionExecutionAsync(ActionExecutingContext context, ActionE
...
+                        await next.Invoke();
+                    }
+                }
+                catch (SecurityTokenExpiredException) {
+                    var unauthorizedResult = new UnauthorizedObjectResult(new ErrorResult("Token expired."));
+                    context.Result = unauthorizedResult;
+                }
+                catch (Exception) {
```
Cepheid-UK, line 41: Need to handle an exception if res is null, it will return 401 and it won't be clear what happened.
If Res = null, we should return 502, which will indicate the issue is in ID service.

gustav87: res won't be null anymore. it will either be a `W3CUserAuthenticationDto` or some exception has been thrown.

PR #328, opened 2024-07-10
Add websocket feature for launcher friend list

gustav87:


W3ChampionsStatisticService/Friends/FriendResponseType.cs
```diff
@@ -0,0 +1,17 @@
+namespace W3ChampionsStatisticService.Friends;
+
+public class FriendResponseType
```
Cepheid-UK, line 3: This seems like more of a job for an enum, if I'm understanding the context correctly you would be better doing something like:

```
public enum FriendResponseType
    {
        [Description("FriendResponseMessage")]
        FriendResponseMessage,

        [Description("FriendResponseData")]
        FriendResponseData,

        [Description("FriendsWithPictures")]
        FriendsWithPictures
    }
    
    FriendResponseType responseType = FriendResponseType.FriendResponseMessage;
    responseType.GetDescription()); // FriendResponseMessage
```

Much more elegant use of the language features, immune to typos and it replaces the `ToString()` function.

Issue #348, opened 2024-10-28
Matches page queries aren't being handled correctly


francislavoie:
Repro steps:
- Go to https://w3champions.com/matches
- Choose a map (some active 1v1 map)
- See the live games are correctly filtered by the map
- Choose "Finished"
- Notice the results contain all maps, not filtered by the one you picked

The problem seems to be with this endpoint:

https://github.com/w3champions/website-backend/blob/ada2d1c5956694a2a2e971f47891d39844c7b760/W3ChampionsStatisticService/Matches/MatchesController.cs#L17-L33

The frontend makes requests like https://website-backend.w3champions.com/api/matches?offset=0&gateway=20&pageSize=50&gameMode=1&map=2404091839ConcealedHill&season=20, i.e. `&map=2404091839ConcealedHill` but this param isn't being read at all by this API endpoint, so the data returned is for all maps instead.

The `gateway` query param isn't being read either, but that probably matters less.

CoffMakesThings:
Fixed by https://github.com/w3champions/website-backend/pull/513

PR #349, opened 2024-10-28
add missing filtering to finished games for #348

gremlinhunter:
Adds missing filters to Finished Games filtering as noted in the following issue:
https://github.com/w3champions/website-backend/issues/348

Unsure as to what refactoring rules there are but some could be applied in MatchesController like the following:

- [unecessary fields due to primary constructors](https://github.com/w3champions/website-backend/blob/master/W3ChampionsStatisticService/Matches/MatchesController.cs#L14-L15) 
- typos such as ["Match" -> "Matche"](https://github.com/w3champions/website-backend/blob/master/W3ChampionsStatisticService/Matches/MatchesController.cs#L51)


MichaelDeitz:
@gremlinhunter  I think this is finished. Can you confirm if this can be closed?

PR #355, opened 2025-01-27
use Directory.Build.Props and Directory.Packages.Props

StojanovicMilos:
This PR introduces two new configuration files to streamline project management:

Directory.Build.props: Centralized common build settings (such as version numbers and build options) for all projects in the directory and subdirectories, ensuring consistency across the solution.

Directory.Packages.props: Introduced to define consistent NuGet package versions across multiple projects, preventing version mismatches and making dependency management easier.

These changes simplify project maintenance, reduce duplication of settings across individual project files and prevents package version conflicts.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+root = true
+
+[*]
+charset = utf-8
+tab_width = 4
+indent_size = 2
+ident_style = space
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,6 +1,6 @@
 root = true
 
-[*]
+[*.csproj]
 charset = utf-8
 tab_width = 4
 indent_size = 2
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,12 +1,6 @@
 root = true
 
-[*.csproj]
-charset = utf-8
-tab_width = 4
-indent_size = 2
-ident_style = space
-
-[*.props]
+[*.{csproj,props}]
 charset = utf-8
 tab_width = 4
 indent_size = 2
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,3 +5,9 @@ charset = utf-8
 tab_width = 4
 indent_size = 2
 ident_style = space
+
+[*.props]
+charset = utf-8
+tab_width = 4
+indent_size = 2
+ident_style = space
--- a/.editorconfig
+++ b/.editorconfig
@@ -4,4 +4,4 @@ root = true
 charset = utf-8
 tab_width = 4
 indent_size = 2
-ident_style = space
+indent_style = space
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,6 +2,5 @@ root = true
 
 [*.{csproj,props}]
 charset = utf-8
-tab_width = 4
 indent_size = 2
 indent_style = space
--- a/W3C.Domain/W3C.Domain.csproj
+++ b/W3C.Domain/W3C.Domain.csproj
@@ -1,14 +1,14 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
-	<ItemGroup>
-		<PackageReference Include="Microsoft.Extensions.Http" />
-		<PackageReference Include="MongoDB.Driver" />
-		<PackageReference Include="Serilog" />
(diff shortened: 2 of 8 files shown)


W3C.Domain/W3C.Domain.csproj
```diff
@@ -1,14 +1,10 @@
...
-    <TargetFramework>net8.0</TargetFramework>
-  </PropertyGroup>
-
+	
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Http" Version="6.0.0" />
-    <PackageReference Include="MongoDB.Driver" Version="2.25.0" />
-    <PackageReference Include="Serilog" Version="2.12.0" />
-    <PackageReference Include="Serilog.Sinks.File" Version="5.0.0" />
+	  <PackageReference Include="Microsoft.Extensions.Http" />
```
marcoabreu, line 4: Any particular reason you're unpinning all the versions?

marcoabreu (approved):

gustav87:
Thanks, looks good, but would you mind checking the indentation?

StojanovicMilos:
Of course, I've formatted all the changed documents with default visual studio spacing

gustav87:
Much better! However, we're not using tabs anywhere else in the project. Could you replace with spaces?

gustav87:
I'm going to add a pipeline step to check formatting, but it does not cover csproj files, so it will have to be manual. Luckily, csproj files are rarely changed. https://github.com/dotnet/format/issues/647

StojanovicMilos:
Ah, apologies, my oversight. To prevent this from happening in the future, I've added .editorconfig file with identation/tabulation rules so that auto-format is uniform for everyone.

gustav87:
Now you have a combination of tabs and spaces in the csproj files.

gustav87:
Looks good. Question, what do we need the .editorconfig file for? I can't get it working properly in vscode. Does it do the job in visual studio? I'm asking because your previous commit was still malformed, and also you have a spelling error in it. `ident_style = space` instead of `indent_style = space`. I would probably suggest deleting the file.

StojanovicMilos:
I had a typo in my editorconfig, that's why it wasn't working. It should be good now.

An .editorconfig file is a configuration file used to define and maintain consistent coding styles in a project, no matter which text editor or IDE you're using. It allows you to set rules like:

How many spaces to use for indentation (2 spaces, 4 spaces, or a tab).
Whether to use spaces or tabs for indentation.
Line endings (Unix-style or Windows-style).
Character encoding.

This helps everyone working on the project follow the same style, even if they're using different tools, and keeps the code clean and consistent. 

Quick google search says it is usable in vsc too, but it looks like you need an extension, this looks like a good place to start: https://github.com/editorconfig/editorconfig-vscode. I've never used it in vsc, but it works wonders in visual studio. Give it a go and let me know if you need any help.

As for removing .editorconfig, I'd suggest against that, because, next to what I've stated above, it can also have configurations on how certain rules in code are treated: none, suggestion, warning, error. For example, if we add rule like this, compiling will not show warning on an error when we have unreachable code due to condition:

dotnet_diagnostic.CA1508.severity = none

However, if we say 

dotnet_diagnostic.CA1508.severity = error

then the code will not compile when there's an unreachable conditional code. You can find the documentation about about this concrete rule: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1508

The idea of .editorconfig is to fill it up with rule definitions such as this one to automate and standardize our compilation and code quality as much as possible, without having to manually check it.

gustav87:
It didn't work, even with the extension, but I see now `dotnet format` picks up the config from the `.editorconfig` file. That's cool! Approved.

gustav87:
`dotnet format` is what I intend to use as a formatting checker in the build pipeline, so that works out well.

.editorconfig
```diff
@@ -0,0 +1,7 @@
+root = true
+
+[*.{csproj,props}]
+charset = utf-8
+tab_width = 4
```
gustav87, line 5:
```suggestion
```

gustav87: Quoted from the editorconfig docs:

```
tab_width: a whole number defining the number of columns used to represent a tab character. This defaults to the value of indent_size and doesn't usually need to be specified.
```

PR #364, opened 2025-03-07
Cache /api/matches/search

CoffMakesThings:
Recently we experienced some website slowness. /api/matches/search is heavily used and does an expensive db query.

My suggestion is to create a MatchService, similar to other services, which leverages our existing in-memory cache. This will cache the results of unique /api/matches/search requests for 15 seconds.

How do I know the caching works? I used the console logs which are removed in the second commit.

Screenshot 2025-03-07 172157
Screenshot 2025-03-07 172223

### DIFF
--- a/W3ChampionsStatisticService/Cache/ICachedDataProvider.cs
+++ b/W3ChampionsStatisticService/Cache/ICachedDataProvider.cs
@@ -5,5 +5,5 @@ namespace W3ChampionsStatisticService.Cache;
 
 public interface ICachedDataProvider<T>
 {
-    Task<T> GetCachedOrRequestAsync(Func<Task<T>> requestDataCallbackAsync, string key);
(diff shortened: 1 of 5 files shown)


marcoabreu (approved):
Lgtm

Whats the cache eviction policy? Like, when do entries get purged from the cache so it doesn't grow forever?

CoffMakesThings:
> Lgtm
> 
> Whats the cache eviction policy? Like, when do entries get purged from the cache so it doesn't grow forever?

In this case this is a 15 second absolute expiration, so after 15 seconds the entry is evicted. I read about absolute expiration here.

https://learn.microsoft.com/en-us/aspnet/core/performance/caching/memory?view=aspnetcore-8.0

godrien (approved):

gustav87 (approved):

PR #369, opened 2025-04-04
Do not create personal setting on get

gustav87:
This stops bogus entries from being added to the personal settings collection.

Frontend PR: https://github.com/w3champions/website/pull/878
image


francislavoie:
Oh wow 😳 that explains a lot of the weird search results we've been getting. Any way we can cull the garbage from the DB afterwards?

W3ChampionsStatisticService/Rewards/Portraits/PortraitCommandHandler.cs
```diff
@@ -8,21 +8,14 @@ namespace W3ChampionsStatisticService.Rewards.Portraits;
...
-    private readonly IPlayerRepository _playerRepository = playerRepository;
     private readonly IPortraitRepository _portraitRepository = portraitRepository;
 
     public async Task<bool> UpdatePicture(string battleTag, SetPictureCommand command)
     {
         var setting = await _personalSettingsRepository.Load(battleTag);
-        if (setting == null)
-        {
-            var playerProfile = await _playerRepository.LoadPlayerProfile(battleTag);
-            setting = new PersonalSetting(battleTag, playerProfile);
```
marcoabreu, line 24: So how are you now making sure that setting is never null?

W3ChampionsStatisticService/PersonalSettings/PersonalSettingsController.cs
```diff
@@ -11,25 +11,18 @@ namespace W3ChampionsStatisticService.PersonalSettings;
...
         {
-            var setting = await _personalSettingsRepository.Load(battleTag);
-            if (setting == null)
-            {
-                var player = await _playerRepository.LoadPlayerProfile(battleTag);
-                return Ok(new PersonalSetting(battleTag) { RaceWins = player });
-            }
-            return Ok(setting);
+            PersonalSetting setting = await _personalSettingsRepository.Load(battleTag, false);
+            return setting == null ? NotFound($"Personal settings of {battleTag} not found.") : Ok(setting);
```
marcoabreu, line 25: The previous code actually initialized the settings, instead you're now throwing an error.

So what is the actually remaining cold-boot path of a new player without settings?

gustav87: It depends on what you're doing first, I think. Logging into the launcher? Logging into the website?

gustav87: Because personal settings are created when GET'ing the api/personal-settings/{battletag} endpoint, and that's needed when visiting a profile for the first time (for example when you log in on the website for the first time, and you're redirected to your profile), I've added that back.

However, if there are no personal settings in the db already created, I ask identity service if the user is valid. If yes, they are created, otherwise I return a 404. This way I don't have to rebuild too much, and we still get protection from bogus personal settings being created.

ID PR: https://github.com/w3champions/identification-service/pull/48

W3ChampionsStatisticService/PersonalSettings/PersonalSettingsController.cs
```diff
@@ -11,23 +12,27 @@ namespace W3ChampionsStatisticService.PersonalSettings;
...
     {
         try
         {
-            var setting = await _personalSettingsRepository.Load(battleTag);
+            PersonalSetting setting = await _personalSettingsRepository.Load(battleTag);
             if (setting == null)
             {
-                var player = await _playerRepository.LoadPlayerProfile(battleTag);
-                return Ok(new PersonalSetting(battleTag) { RaceWins = player });
+                bool userExists = await _identityServiceClient.UserExists(battleTag);
```
marcoabreu, line 30: Yeah I like this approach, thanks!

W3ChampionsStatisticService/PlayerProfiles/PlayersController.cs
```diff
@@ -40,49 +40,24 @@ public async Task<IActionResult> GlobalSearchPlayer(string search, string lastRe
...
-            }
+            return NotFound($"Player {battleTag} not found.");
         }
 
         // Akas are stored in cache - preferences for showing akas are stored in DB
-        var settings = await _personalSettingsRepository.Load(battleTag);
+        PersonalSetting settings = await _personalSettingsRepository.LoadOrCreate(battleTag);
         player.PlayerAkaData = await _playerAkaProvider.GetAkaDataByPreferencesAsync(battleTag, settings);
 
         await _playerRepository.UpsertPlayer(player);
```
marcoabreu, line 56: Why are we updating every time we get a player?

marcoabreu (approved):

W3ChampionsStatisticService/PlayerProfiles/PlayersController.cs
francislavoie: The repository does nothing if the element hasn't actually changed, pretty sure.

W3ChampionsStatisticService/PersonalSettings/PersonalSettingsRepository.cs
```diff
@@ -14,16 +14,38 @@ public class PersonalSettingsRepository(MongoClient mongoClient) : MongoDbReposi
...
     {
-        var personalSettings = await LoadFirst<PersonalSetting>(battletag);
+        PersonalSetting personalSettings = await LoadFirst<PersonalSetting>(battletag);
+
+        if (personalSettings == null)
+        {
+            return null;
+        }
+
+        PlayerOverallStats playerStats = await CreateCollection<PlayerOverallStats>()
```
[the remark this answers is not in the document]
gustav87: I think it only makes the collection available for querying, I'm not sure. I'm not doing that differently than before, I just added another method Load beside LoadOrCreate, which does not create personal settings when loading and not finding.

PR #374, opened 2025-04-14
Matchup: Add finished match player heroes to matchup response data

Magentah:
Adds `Hero` info to the `Matchup` object so it can be returned to the front-end.

Implementation is a little awkward because of the various different types for player data but I wanted to make minimal changes until I get more familiar with things.

### DIFF
--- a/W3ChampionsStatisticService/Matches/PlayerOverviewMatches.cs
+++ b/W3ChampionsStatisticService/Matches/PlayerOverviewMatches.cs
@@ -1,4 +1,6 @@
-﻿using W3C.Contracts.GameObjects;
+﻿using System.Collections.Generic;
+using W3C.Contracts.GameObjects;
+using W3C.Domain.MatchmakingService;
 
 namespace W3ChampionsStatisticService.Matches;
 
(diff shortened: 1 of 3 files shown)


francislavoie:
Thanks! (Ruhigengeist here)

Since this is working off the match finished event, I think this would only add the data for new matches, right? Unless we replay the events I guess.

(Not my call if this gets merged, hopefully others agree this is a good idea! 😅)

Magentah:
It would only add data for new matches, yes. I'm not sure how data updates / migrations work so I didn't do that yet

W3ChampionsStatisticService/Matches/PlayerOverviewMatches.cs
```diff
@@ -17,4 +19,5 @@ public class PlayerOverviewMatches
     public string CountryCode { get; set; }
     public string Country { get; set; }
     public string Twitch { get; set; }
+    public List<Hero> Heroes { get; set; }
```
marcoabreu, line 22: Shall we make it `List<Hero>? Heros`

Magentah:
Closing as there's a separate PR

PR #380, opened 2025-04-27
Add Hero data and filtering to Matchup

Magentah:
- Add new `Hero` class that specifically for backend to separate from the MM MatchFinishedEvent hero class. This also parses the `icon` path differently. The parsing can be expanded to support custom games later if we want
- Add new `HeroType` enum that's used as an ID for heroes instead of using strings. The enum value is saved in the `Matchup` document and is used for querying and filtering.
- Add hero data for players in `Matchup`. This only adds the enum value as an id, and the level. Strings are handled in the backend/frontend as needed.
- Add backfill service to gradually update historical matchup data to add hero data. This runs as a background task updating in batches of 1000 with a 5 second delay between each batch (set for all `ReadModelHandler` background tasks). Is enabled by setting the `RUN_BACKFILL=true` environment variable.
- Updates `api/matches` to accept hero ids for filtering. 
- Updates `PlayerScore` and `PlayerOverviewMatches` to use new `Hero` class for consistency.

A new index needs to be added to `Matchup.Teams.Players.Heroes._id` otherwise filtering is going to be very slow.

### DIFF
--- a/W3ChampionsStatisticService/Hero/Hero.cs
+++ b/W3ChampionsStatisticService/Hero/Hero.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System.Text.Json.Serialization;
@@ -20,7 +21,7 @@ public class Hero
     public HeroType Id { get; set; }
 
     [BsonIgnore()]
-    public string Name => Enum.GetName(Id)?.ToLower(); // Frontend uses lowercase for translation keys
+    public string Name { get; private set; }
 
     /// <summary>
     /// For backwards compatibility to not break the frontend
@@ -31,10 +32,25 @@ public class Hero
 
     public int Level { get; set; }
 
+    // Don't need to store in MongoDB cos can infer from Id values. Whether hero data is for classic, reforged, custom, etc
+    [BsonIgnore()]
+    public HeroSource Source { get; private set; }
+
     public Hero(W3C.Domain.MatchmakingService.Hero heroData)
     {
         Id = ParseHeroIcon(heroData.iconPath);
         Level = heroData.level;
+        Source = ParseHeroSource(Id);
+        Name = ParseHeroName(Id, Source);
+    }
+
+    [BsonConstructor("Id", "Level")]
+    public Hero(HeroType type, int level)
+    {
+        Id = type;
+        Level = level;
+        Source = ParseHeroSource(Id);
+        Name = ParseHeroName(Id, Source);
     }
 
     /// <summary>
@@ -61,10 +77,6 @@ public class Hero
         {
             return (HeroType)parsedHeroId;
         }
-        else if (Enum.TryParse(typeof(ReforgedHeroType), heroName, true, out var parsedReforgedId))
-        {
-            return MapReforged((ReforgedHeroType)parsedReforgedId);
-        }
         else
         {
             // TODO: Add custom game mapping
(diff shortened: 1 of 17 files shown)


W3ChampionsStatisticService/Matches/MatchRepository.cs
```diff
@@ -176,12 +175,24 @@ public async Task<int> GetFloIdFromId(string gameId)
...
+    private FilterDefinition<Matchup> GetLoadFilter(int season, GameMode gameMode, HeroType hero = HeroType.AllFilter)
     {
-        return CreateCollection<Matchup>().CountDocumentsAsync(m =>
-                gameMode == m.GameMode && m.Season == season);
+        var builder = Builders<Matchup>.Filter;
+        var filter = builder.Eq(m => m.GameMode, gameMode) & builder.Eq(m => m.Season, season);
+
+        if (hero != HeroType.AllFilter && hero != HeroType.Unknown)
+        {
+            var heroFilter = builder.Where(m => m.Teams.Any(t => t.Players.Any(p => p.Heroes.Any(h => h.Id == hero))));
```
francislavoie, line 191: I think it should probably match only the first hero in the list (vast majority of the time you'd want to search for opening hero for build order stuff, I think) but I'm not sure if the game is reliable about keeping the list in order they were trained

Magentah: Yeah, I was thinking about that. Not sure if the list we get is in order or not, but I guess we can just assume it is for now and see how it goes. Could also filter on highest level but would need adding an index on that as well I think which probably isn't worth it

Magentah:
Draft so I can add tests and fix builds

W3ChampionsStatisticService/Hero/HeroController.cs
```diff
@@ -0,0 +1,22 @@
+using System;
+using System.Linq;
+using Microsoft.AspNetCore.Mvc;
+
+namespace W3ChampionsStatisticService.Heroes;
+
+[ApiController]
+[Route("api/hero")]
```
marcoabreu, line 8: Can you ensure that this response is a long cache duration?

W3ChampionsStatisticService/Matches/MatchRepository.cs
```diff
@@ -148,25 +152,20 @@ private static PlayerScore CreateDetail(PlayerBlizzard playerBlizzard)
...
         var mongoCollection = CreateCollection<Matchup>();
-        return mongoCollection
-            .Find(m => gameMode == m.GameMode && m.Season == season)
-            .SortByDescending(s => s.EndTime)
-            .Skip(offset)
-            .Limit(pageSize)
-            .ToListAsync();
+        var filter = GetLoadFilter(season, gameMode, hero);
+
+        var results = mongoCollection.Find(filter).SortByDescending(s => s.EndTime).Skip(offset).Limit(pageSize).ToListAsync();
```
marcoabreu, line 166: Did you double check that the index order is in the same direction as your sort?

W3ChampionsStatisticService/Matches/MatchRepository.cs
```diff
@@ -176,12 +175,24 @@ public async Task<int> GetFloIdFromId(string gameId)
...
+    private FilterDefinition<Matchup> GetLoadFilter(int season, GameMode gameMode, HeroType hero = HeroType.AllFilter)
     {
-        return CreateCollection<Matchup>().CountDocumentsAsync(m =>
-                gameMode == m.GameMode && m.Season == season);
+        var builder = Builders<Matchup>.Filter;
+        var filter = builder.Eq(m => m.GameMode, gameMode) & builder.Eq(m => m.Season, season);
+
+        if (hero != HeroType.AllFilter && hero != HeroType.Unknown)
+        {
+            var heroFilter = builder.Where(m => m.Teams.Any(t => t.Players.Any(p => p.Heroes.Count > 0 && p.Heroes[0].Id == hero)));
```
marcoabreu, line 191: How come you're only checking Heroes[0]?

W3ChampionsStatisticService/Matches/MatchRepository.cs
Magentah: I don't have access to the database so I don't know what indexes exist. The public dump is missing indexes. Either way, this starts from the most recent which, I assume, is what we would want by default anyway as sorting from the oldest first makes no sense.

W3ChampionsStatisticService/Matches/MatchRepository.cs
Magentah: Suggestion from @francislavoie to only filter on the first hero instead of all heroes

W3ChampionsStatisticService/Hero/HeroController.cs
Magentah: Done for 7 days, seems reasonable enough

W3ChampionsStatisticService/Matches/MatchRepository.cs
marcoabreu: The indizes are defined in the code: https://github.com/w3champions/website-backend/blob/9c4040b8daa556e7b84145a2527bf574f9e31d36/W3ChampionsStatisticService/Matches/MatchRepository.cs#L125

W3ChampionsStatisticService/Matches/MatchRepository.cs
marcoabreu: Hmm okay, feels odd that we're only selecting the first hero, but well

W3ChampionsStatisticService/Matches/MatchRepository.cs
marcoabreu: But here's a screeenshot:
image

W3ChampionsStatisticService/Matches/MatchRepository.cs
francislavoie: Well it's more about "what's the first hero in the build order". Getting MK first is not the same thing as being it second or third, that's a different build entirely, so filtering on games with that too wouldn't do much. Think about it, a human player might alternate between 2 or 3 different strategies, but all of the strats could get the same one hero eventually (except for fast games of course that end with one hero).

W3ChampionsStatisticService/Matches/MatchRepository.cs
Magentah: That's only the text index for `Team1-4Players` fields. The `Season, GameMode, EndTime` index has EndTime in descending though, which this would be using as it includes Season and GameMode as well.

W3ChampionsStatisticService/Hero/Hero.cs
```diff
@@ -0,0 +1,146 @@
...
+            case ReforgedHeroType.JainaSea:
+                return HeroType.Archmage;
+            case ReforgedHeroType.ThrallChampion:
+                return HeroType.Farseer;
+            case ReforgedHeroType.FallenKingArthas:
+                return HeroType.DeathKnight;
+            case ReforgedHeroType.CenariusNightmare:
+                return HeroType.KeeperOfTheGrove;
+            default:
+                return HeroType.Unknown;
```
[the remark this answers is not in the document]
Magentah: I can add versioning of some kind if we want by using [Vary](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Vary)?

W3ChampionsStatisticService/Hero/HeroController.cs
Magentah: I can add versioning of some kind using [Vary](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Vary?

Magentah: Updated to include `Vary` for the response cache using the hashcode of the enum.

PR #388, opened 2025-05-06
Increase player global search page size limit to 100

francislavoie:
I'm looking to use this API in `launcher-e`, but since infinite scroll will probably be difficult, I'd rather have a higher page size limit. I think the cap of 20 is kinda low, I think it should be okay to get up to 100 results in one call if the caller wants it. Default remains at 20, only changing the `&pageSize=` allowed maximum. Planning to use 50 in `launcher-e`.

### DIFF
--- a/W3ChampionsStatisticService/PlayerProfiles/PlayersController.cs
+++ b/W3ChampionsStatisticService/PlayerProfiles/PlayersController.cs
@@ -36,7 +36,7 @@ public class PlayersController(
     [HttpGet("global-search")]
     public async Task<IActionResult> GlobalSearchPlayer(string search, string lastRelevanceId = "", int pageSize = 20)
     {
-        if (pageSize > 20) pageSize = 20;
+        if (pageSize > 100) pageSize = 100;
         var players = await _playerService.GlobalSearchForPlayer(search, lastRelevanceId, pageSize);
         return Ok(players);
     }


marcoabreu (changes requested):
I am strongly against that. Pagination exists for a reason and this creates unnecessary load.

marcoabreu:
Please elaborate why lazy loading should be hard in the launcher. This will create a fivefold amount of database and network traffic that gets kicked on every single character someone types in.

francislavoie:
Infinite scrolling is a non-trivial piece of functionality. Primereact doesn't seem to support it in a sensible way (all kinds of jank with their implementation that I think is unacceptable).

I don't see the point of a `pageSize` argument at all if it can't be increased to a sensible number. There's no reason to ever choose a number lower than 20, so why even have it as a param, if we can't raise it when we need to?

I don't see the problem with allowing up to 100, why would that be a problem? 100 records would not make a measurable impact on the API, especially since vast majority of searches will result in much less than that anyway. Why do more round-trips to the server, adding _more_ overhead than just getting as much relevant data up-front in one shot?

marcoabreu:
Because 20 is a sensible number and 100 is not. You're retrieving up to 100 items for every single character a user enters, so your assumption that it will only be few results is not held true as someone types. 

Here's some code to achieve lazy loading without the overlay - feel free to adapt accordingly. 

It has a spinner at the bottom but won't hide the already down data. 

```
import React, { useState, useEffect } from 'react';
import { DataTable } from 'primereact/datatable';
import { Column } from 'primereact/column';
import { ProgressSpinner } from 'primereact/progressspinner';

const MyInfiniteScrollTable = () => {
  const [data, setData] = useState([]);
  const [lazyLoading, setLazyLoading] = useState(false);
  const [totalRecords] = useState(1000); // total items available from server
  const [loadedRows, setLoadedRows] = useState(0);
  const rowsPerLoad = 20;

  const loadMore = (event) => {
    if (lazyLoading) return;
    setLazyLoading(true);

    // Simulate async fetch
    setTimeout(() => {
      const newItems = Array.from({ length: rowsPerLoad }, (_, i) => ({
        id: loadedRows + i + 1,
        name: `Item #${loadedRows + i + 1}`
      }));
      setData((prev) => [...prev, ...newItems]);
      setLoadedRows((prev) => prev + rowsPerLoad);
      setLazyLoading(false);
    }, 1000);
  };

  // Optional: load initial data
  useEffect(() => {
    loadMore({});
  }, []);

  const loadingTemplate = lazyLoading ? (
    <div style={{ textAlign: 'center', padding: '1rem' }}>
      <ProgressSpinner style={{ width: '30px', height: '30px' }} strokeWidth="4" />
    </div>
  ) : null;

  return (
    <DataTable
      value={data}
      scrollable
      scrollHeight="400px"
      lazy
      onVirtualScroll={loadMore}
      virtualScroll
      rows={rowsPerLoad}
      totalRecords={totalRecords}
      loading={false} // avoid overlay
      loadingIcon={false}
      virtualScrollDelay={0}
      footer={loadingTemplate}
    >
      <Column field="id" header="ID" />
      <Column field="name" header="Name" />
    </DataTable>
  );
};

export default MyInfiniteScrollTable;
```

francislavoie:
> You're retrieving up to 100 items for every single character a user enters, so your assumption that it will only be few results is not held true as someone types.

Actually there's debouncing, and it only searches if you've given at least 3 characters, so it's really not that bad. There's not a lot of 3-letter combinations that will return more than like 50 results anyway.

Anyway, I'll close this cause we decided we'll implement infinite scrolling in launcher-e, but I still think it's exceptionally weird that we set the maximum the same as the default, and I think the upper allowable limit is much too low. Deciding for the frontend that the maximum page size is 20 is weird, I find it very common in many apps to have a user-selectable "page size" option (especially data tables) and a page size of 100 max is pretty typical for that kind of pattern.

PR #389, opened 2025-05-07
Update websockets to include friend online status

Magentah:
- Update `FriendUser` to include `IsOnline` property.
- Update `ConnectionMapping` to also keep a mapping of `BattleTag, HashSet<ConnectionId>` so we don't need to search constantly to check for whether a user is online or not.
- Update `ConnectionMapping` to add `IsUserOnline()` to check whether a user is online or not.
- Update `WebsiteBackendHub#GetFriends()` to get the online status of users using the `ConnectionMapping#IsUserOnline()` from above

### DIFF
--- a/W3ChampionsStatisticService/Hubs/ConnectionMapping.cs
+++ b/W3ChampionsStatisticService/Hubs/ConnectionMapping.cs
@@ -6,10 +6,12 @@ namespace W3ChampionsStatisticService.Hubs;
 public class ConnectionMapping
 {
     private readonly Dictionary<string, WebSocketUser> _connections = new();
+    private readonly Dictionary<string, HashSet<string>> _onlineUsersByBattleTag = new();
+    private readonly object _lock = new();
 
     public List<WebSocketUser> GetUsers()
     {
-        lock (_connections)
+        lock (_lock)
         {
             return _connections.Values.Select(v => v).OrderBy(r => r.BattleTag).ToList();
         }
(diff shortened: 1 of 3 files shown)


W3ChampionsStatisticService/Hubs/ConnectionMapping.cs
```diff
@@ -6,40 +6,67 @@ namespace W3ChampionsStatisticService.Hubs;
...
+                }
+                _onlineUsersByBattleTag[user.BattleTag].Add(connectionId);
             }
         }
     }
 
     public WebSocketUser GetUser(string connectionId)
     {
-        lock (_connections)
+        lock (_lock)
```
marcoabreu, line 38: Hmm, locking on every get seems tough. Could you check if we have an easy way to get a read write lock?

marcoabreu:
Another thought: can we do a bulk check where you give in a list of battle tags and it just does an intersect. Rather than doing the IsOnline call for each item individually?
Avoids X sequential lookups.

W3ChampionsStatisticService/Hubs/ConnectionMapping.cs
Magentah: Updated

Magentah:
> Another thought: can we do a bulk check where you give in a list of battle tags and it just does an intersect. Rather than doing the IsOnline call for each item individually? Avoids X sequential lookups.

The lookup is O(1), It's just checking whether a key exists. I doubt it's any faster or better to do an intersect

francislavoie:
I think it's more about each sequential call locking and unlocking being expensive, if you can group it all in one call, one lock, then it should be better, right?

marcoabreu (approved):
Lgtm, thanks a lot!

PR #390, opened 2025-05-08
Send user info on connect and disconnect to friends

Magentah:
Previously was only sending the boolean which isn't useful. Updates to send the battle tag as well so it's actually useful


francislavoie (approved):
Thanks!

PR #391, opened 2025-05-09
Introduce transaction support, rewrite cache and introduce match cleanup

marcoabreu:
This PR introduces transactions to the website-backend service in order to ensure consistent execution of our operations.

The matchup cache has been rewritten to ensure good performance.

Match canceled events are now properly consumed.

Debends on database PR: https://github.com/w3champions/docker-compose-files/pull/71

### DIFF
--- a/W3C.Domain/Repositories/MongoDbRepositoryBase.cs
+++ b/W3C.Domain/Repositories/MongoDbRepositoryBase.cs
@@ -2,8 +2,11 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Linq.Expressions;
+using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.VisualBasic;
 using MongoDB.Driver;
+using MongoDB.Driver.Linq;
 
 namespace W3C.Domain.Repositories;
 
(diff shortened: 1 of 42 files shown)


W3C.Domain/Repositories/MongoDbRepositoryBase.cs
```diff
@@ -37,15 +42,28 @@ protected Task<T> LoadFirst<T>(string id) where T : IIdentifiable
...
+        var session = _transactionCoordinator?.GetCurrentSession();
+        return session != null
+            ? mongoCollection.InsertOneAsync(session, element)
+            : mongoCollection.InsertOneAsync(element);
     }
 
     protected async Task<List<T>> LoadAll<T>(Expression<Func<T, bool>> expression = null, int? limit = null)
     {
-        if (expression == null) expression = l => true;
+        expression ??= l => true;
```
Magentah, line 53: If there's no expression, it's probably better to just find with an empty filter instead? It's at least more readable and may be handled by mongo better?

```csharp
var filter = Builders<T>.Filter.Empty;
mongoCollection.Find(filter);
```

W3C.Domain/Repositories/MongoDbRepositoryBase.cs
```diff
@@ -62,12 +80,18 @@ protected IMongoCollection<T> CreateCollection<T>(string collectionName = null)
...
-        var mongoCollection = mongoDatabase.GetCollection<T>(typeof(T).Name);
-        await mongoCollection.FindOneAndReplaceAsync(
-            identityQuerry,
-            insertObject,
-            new FindOneAndReplaceOptions<T> { IsUpsert = true });
+        var mongoCollection = CreateCollection<T>();
+        var session = _transactionCoordinator?.GetCurrentSession();
+        var options = new FindOneAndReplaceOptions<T> { IsUpsert = true };
+
+        if (session != null)
```
Magentah, line 87: Maybe good to be consistent with using ternary or not. 40 lines up a ternary is used for this exact same thing

W3C.Domain/Repositories/MongoDbTransactionCoordinator.cs
```diff
@@ -0,0 +1,103 @@
...
+
+    public MongoDbTransactionCoordinator(IMongoClient mongoClient)
+    {
+        _mongoClient = mongoClient ?? throw new ArgumentNullException(nameof(mongoClient));
+        _onSuccessHandlers = new ConcurrentQueue<Func<Task>>();
+    }
+
+    public async Task BeginTransactionAsync(CancellationToken cancellationToken = default)
+    {
+        if (IsTransactionActive)
```
Magentah, line 27: Does it make sense that we can only ever have a single request happening at once per collection, otherwise we throw errors? Would it not be better to instead have multiple sessions and let MongoDB deal with ordering transactions as needed? 

It seems to me that throwing exceptions whenever there's more than 1 request at a time is probably not ideal.

W3ChampionsStatisticService/Matches/CanceledMatchesHandler.cs
```diff
@@ -0,0 +1,74 @@
...
+                        {
+                            var ongoingMatch = await _matchRepository.LoadDetailsByOngoingMatchId(nextEvent.match.id);
+
+                            if (ongoingMatch == null)
+                            {
+                                Log.Warning($"Canceled match {nextEvent.match.id} not found");
+                            }
+                            else
+                            {
+                                if (ongoingMatch.Match != null)
```
Magentah, line 47: Let's not have 7 nested levels of logic

W3C.Domain/Repositories/MongoDbRepositoryBase.cs
marcoabreu: Good call, thanks!

W3C.Domain/Repositories/MongoDbTransactionCoordinator.cs
marcoabreu: Oh boy, you're right! At some point I had the session variable within an AsyncLocal but somehow the contexts got messed up and I removed it again. That's why I had this who IsTransactionActive be a dynamic call. Yeah, that's a very good call, thank you!

W3ChampionsStatisticService/Matches/CanceledMatchesHandler.cs
marcoabreu: I tried to keep the original logic of the other handlers to keep the diff "small".

Happy to rewrite all of them.

W3ChampionsStatisticService/Matches/OngoingMatchesCache.cs
```diff
@@ -20,12 +92,28 @@ public async Task<long> CountOnGoingMatches(
...
+        {
+            if (cachedValue.IsNegativeCache)
+            {
+                return 0L;
+            }
+
+            return cachedValue.Value;
+        }
+
+        await EnsureCachesInitialized();
```
Magentah, line 106: Why are we ensuring cache initialization after reading from a cache? I know it doesn't init the count cache, but it still reads weirdly

marcoabreu: Yeah it was more about keeping the read and the write paths separate, but I can see that one might want to have it all the start. I don't really like th seeding of the cache anyway tbh :/

Would like to have a proper init method :/

W3ChampionsStatisticService/Matches/OngoingMatchesCache.cs
```diff
@@ -38,24 +126,46 @@ public async Task<List<OnGoingMatchup>> LoadOnGoingMatches(
         int maxMmr,
         string sort)
```
Magentah, line 127: If we're changing things anyway, I think an enum would be better than using a string for this, but I dunno if we want to do that right now

W3ChampionsStatisticService/Matches/OngoingMatchesCache.cs
```diff
@@ -38,24 +126,46 @@ public async Task<List<OnGoingMatchup>> LoadOnGoingMatches(
...
+            matches = _ongoingMatchesCache.Values
+                .Where(m => (gameMode == GameMode.Undefined || m.GameMode == gameMode)
+                            && (gateWay == GateWay.Undefined || m.GateWay == gateWay)
+                            && (map == "Overall" || m.Map == map)
+                            && (minMmr == 0 || !m.Teams.Any(team => team.Players.Any(player => player.OldMmr < minMmr)))
+                            && (maxMmr == 3000 || !m.Teams.Any(team => team.Players.Any(player => player.OldMmr > maxMmr))));
+
+            matches = sort switch
+            {
+                "mmrDescending" => matches.OrderByDescending(GetMaxMmrInMatch),
```
Magentah, line 153: What about ascending? I assume they're valid if given in the API, even if they're technically the default when not given any values?

W3ChampionsStatisticService/Matches/OngoingMatchesHandler.cs
```diff
@@ -24,30 +31,43 @@ public async Task Update()
...
-                            {
-                                player.CountryCode = personalSettings.CountryCode;
-                            }
+                            var matchup = OnGoingMatchup.Create(nextEvent);
+
+                            foreach (var team in matchup.Teams)
+                                foreach (var player in team.Players)
+                                {
+                                    var foundMatchForPlayer = await _matchRepository.LoadOnGoingMatchForPlayer(player.BattleTag);
+                                    if (foundMatchForPlayer != null)
```
Magentah, line 46: Let's also reduce nesting here

W3ChampionsStatisticService/Program.cs
```diff
@@ -88,13 +88,19 @@
 });
 
 // Configure and add MongoDB
-string mongoConnectionString = Environment.GetEnvironmentVariable("MONGO_CONNECTION_STRING") ?? "mongodb://157.90.1.251:3513"; // "mongodb://localhost:27017";
+string mongoConnectionString = Environment.GetEnvironmentVariable("MONGO_CONNECTION_STRING") ?? "mongodb://157.90.1.251:3513?replicaSet=rs0"; // "mongodb://localhost:27017?replicaSet=rs0";
+if (!mongoConnectionString.Contains("replicaSet="))
```
Magentah, line 92: Maybe don't throw an exception for this, just fallback to not using transactions with a warning logged?

W3ChampionsStatisticService/ReadModelBase/ReadModelHandler.cs
```diff
@@ -12,50 +12,56 @@ public class ReadModelHandler<T> : IAsyncUpdatable where T : IReadModelHandler
...
                 {
-                    if (lastVersion.IsStopped) return;
-
-                    if (nextEvent.match.season > lastVersion.Season)
+                    await using (var transaction = await AsyncTransactionScope.CreateAsync(_transactionCoordinator))
                     {
-                        await _versionRepository.SaveLastVersion<T>(lastVersion.Version, nextEvent.match.season);
-                        lastVersion = await _versionRepository.GetLastVersion<T>();
-                    }
+                        if (nextEvent.match.season > lastVersion.Season)
```
Magentah, line 46: Same thing with nesting

W3ChampionsStatisticService/Matches/CanceledMatchesHandler.cs
Magentah: I think if we're making some changes anyway, better to leave things better than we found them. I think it makes sense to improve things if we're already making changes to that code anyway. Can always split things in to separate commits to make PRs easier to review

W3ChampionsStatisticService/Program.cs
marcoabreu: Supporting two parallel modes is just a headache

MichaelDeitz:
@marcoabreu Is this complete?

PR #392, opened 2025-05-10
Add friend list cache. Add friend command handler for simplification

gustav87:


W3ChampionsStatisticService/Hubs/WebsiteBackendHub.cs
```diff
@@ -16,17 +16,19 @@ public class WebsiteBackendHub(
     IW3CAuthenticationService authenticationService,
     ConnectionMapping connections,
     IHttpContextAccessor contextAccessor,
-    FriendRepository friendRepository,
     FriendRequestCache friendRequestCache,
-    IPersonalSettingsRepository personalSettingsRepository
+    IPersonalSettingsRepository personalSettingsRepository,
+    FriendCommandHandler friendCommandHandler,
+    FriendListCache friendListCache
```
Magentah, line 22: I don't think this needs the cache if it can just go through the handler, which uses a cache internally anyway

W3ChampionsStatisticService/Hubs/WebsiteBackendHub.cs
```diff
@@ -54,8 +56,10 @@ public async Task LoadFriendListAndRequests()
     {
         var currentUser = _connections.GetUser(Context.ConnectionId)?.BattleTag;
         if (currentUser == null)
+        {
             return;
-        Friendlist friendList = await _friendRepository.LoadFriendlist(currentUser);
+        }
+        Friendlist friendList = await _friendListCache.LoadFriendList(currentUser);
```
Magentah, line 62: I think this (and the other calls in this file) should be using the handler not the cache?

gustav87: It wouldn't do anything more than call the cache anyway, so there's no need to abstract it.

gustav87: But if you think it makes more sense that way, I'll do it!

gustav87: Scratch that. You're right.

W3ChampionsStatisticService/Hubs/WebsiteBackendHub.cs
gustav87: Yes, thank you!

PR #393, opened 2025-05-10
Add support for canceled matches

marcoabreu:
Process events for canceled matches to remove them from the ongoing matches.

Also rewrites the ongoing matches cache to improve performance.


marcoabreu:
I'll switch the handler to not delete the processed events, in line with the finished one. 

That will allow us to process them for statistics etc.

marcoabreu:
Splitting the PR into further parts

PR #396, opened 2025-05-22
Introduce tracing instrumentation

marcoabreu:
This PR introduces tracing to website-backend. HTTP as well as websocket calls are instrumented and will be traced. In case a tracing context is already present, it will be inherited and hence allow us to do cross-service traces such as this example:
image

We are also propagating the Faro Session ID in order to correlate spans within a session. 

It was necessary to update the dependencies in order to properly instrument everything. I had to do some slight modifications (S3 and Telemetry), but they should all be good from what I can tell.

I did not update nunit to version 4 because they have made some breaking changes that would've required rewriting around 600 assertions.

### DIFF
--- a/W3C.Domain/ChatService/ChatServiceClient.cs
+++ b/W3C.Domain/ChatService/ChatServiceClient.cs
@@ -6,6 +6,7 @@ using System.Text;
 using System.Threading.Tasks;
 using System.Web;
 using W3C.Contracts.Admin.Moderation;
+using W3C.Domain.Tracing;
 
 namespace W3C.Domain.ChatService;
 
@@ -29,6 +30,7 @@ public class ChatServiceClient
         };
     }
 
+    [Trace]
     public async Task<LoungeMuteResponse[]> GetLoungeMutes(string authorization)
     {
         var url = $"{ChatServiceApiUrl}/api/loungeMute/?authorization={authorization}&secret={AdminSecret}";
@@ -42,6 +44,7 @@ public class ChatServiceClient
         return deserializeObject;
     }
 
+    [Trace]
     public async Task<HttpResponseMessage> PostLoungeMute(LoungeMute loungeMute, string authorization)
     {
         var url = $"{ChatServiceApiUrl}/api/loungeMute/?authorization={authorization}&secret={AdminSecret}";
@@ -54,6 +57,7 @@ public class ChatServiceClient
         return response;
     }
 
+    [Trace]
     public async Task<HttpResponseMessage> DeleteLoungeMute(string battleTag, string authorization)
     {
         var encodedTag = HttpUtility.UrlEncode(battleTag);
--- a/W3C.Domain/ChatService/ChatServiceClient.cs
+++ b/W3C.Domain/ChatService/ChatServiceClient.cs
@@ -31,7 +31,7 @@ public class ChatServiceClient
     }
 
     [Trace]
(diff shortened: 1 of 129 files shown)


W3ChampionsStatisticService/Services/Tracing/Interceptors/TracingInterceptor.cs
```diff
@@ -0,0 +1,98 @@
...
+            {
+                ParameterInfo[] parameters = invocation.Method.GetParameters();
+                for (int i = 0; i < invocation.Arguments.Length; i++)
+                {
+                    // Check if the parameter has [NoTrace]
+                    if (parameters[i].GetCustomAttribute<NoTraceAttribute>() == null)
+                    {
+                        var arg = invocation.Arguments[i];
+                        var paramName = parameters[i].Name;
+                        activity.SetTag($"param.{paramName}", arg?.ToString() ?? "null");
```
Magentah, line 74: How does this handle function arguments?

W3C.Domain/ChatService/ChatServiceClient.cs
```diff
@@ -29,6 +30,7 @@ public ChatServiceClient(IHttpClientFactory httpClientFactory)
         };
     }
 
+    [Trace]
```
Magentah, line 33: Probable want `[NoTrace]` on the auth params here?

W3C.Domain/UpdateService/UpdateServiceClient.cs
```diff
@@ -4,9 +4,11 @@
 using System.Net.Http;
 using System.Threading.Tasks;
 using W3C.Domain.UpdateService.Contracts;
+using W3C.Domain.Tracing;
 
 namespace W3C.Domain.UpdateService;
 
+[Trace]
```
Magentah, line 11: Is adding `HttpRequestMessage` to traces potentially problematic? Not sure what the `ToString()` would result in and whether it may include headers or something

W3ChampionsStatisticService/Services/Tracing/Interceptors/TracingInterceptor.cs
marcoabreu: The way it works is it iterates through all function arguments and adds them to the span as "param.searchQuery", for example. The value is always the result of ToString() - unless the NoTrace annotation is set.

W3C.Domain/ChatService/ChatServiceClient.cs
marcoabreu: Oh yeah, good catch. Thanks!

W3C.Domain/UpdateService/UpdateServiceClient.cs
marcoabreu: I will double check!

marcoabreu: image

So it seems like sensitive parameters are automatically redacted and the headers are not traced.

W3ChampionsStatisticService/Services/Tracing/Interceptors/TracingInterceptor.cs
marcoabreu: As clarified via PM, the question was how functions as parameters (to allow dynamic invocations) are handled. The answer is that they aren't, they'll just be converted to string and have no meaningful value or instrumentation.

Can add that if the need arises.

Magentah (approved):

PR #398, opened 2025-05-23
Add global mute search query, nextId and created at properties

gustav87:


W3ChampionsStatisticService/Admin/AdminRepository.cs
```diff
@@ -115,42 +115,42 @@ public async Task<List<string>> SearchSmurfsFor(string tag)
...
-        var content = await result.Content.ReadAsStringAsync();
+        string url = $"{MatchmakingApiUrl}/flo/globalChatBans";
+        if (query != null)
+        {
+            url += $"?query={query}";
+        }
+        if (nextId != null)
+        {
+            url += $"?nextId={nextId}";
+        }
```
francislavoie, line 130: Both have `?`, is that right?

francislavoie: (same on `website` PR btw)

marcoabreu (approved):

W3ChampionsStatisticService/Admin/GlobalChatBan.cs
```diff
@@ -53,4 +56,5 @@ public class ChatBanPutDto
 {
     public string battleTag { get; set; }
     public string expiresAt { get; set; }
+    public string author { get; set; }
```
marcoabreu, line 59: Can we infer the author from the auth data, rather than it being an input field?

PR #409, opened 2025-05-26
Rewrite Ongoing Matches Cache

marcoabreu:
The current caching solution has various issues that this PR is trying to address:
1. *Non-threadsafe data structure*: The List and Linq constructs we are using are not threadsafe. But we are having concurrent read and write operations on these datastructures which can have non-deterministic outcomes.
2. Re-enumeration on every request*. We are filtering and sorting the data on every single request. Considering point 1, we're amplifying the potential for concurrency issues. Aside from that, it is very inefficient.

This PR introduces two types of caches:
1. **ConcurrentDictionary**: This is basically our database lookup cache with either `MatchId` or `PlayerBattleTag` as key. 
2. **MemoryCache**: This is our query cache which stores the actually computed, filtered and ordered results.

The reason two different cache methods were used (I wanted to only use MemoryCache due to its timeout and cache eviction capabilities) is the fact that `MemoryCache` does not permit access to the values. Hence, we would not be able to do the aggregations we need to do.
On the other hand, I didn't want to only use ConcurrentDictionary as this would've meant that we have to manually maintain the cache lifecycle.

We should, in my opinion, evaluate whether we can't just push this down to a database rather than running all this ourselves - databases are hyper-optimized for exactly this use-case. But that can be a task for the future.

I have added a test for heroes in order to validate that we're properly respecting the cache keys.

### DIFF
--- a/W3C.Contracts/Matchmaking/MatchSortMethod.cs
+++ b/W3C.Contracts/Matchmaking/MatchSortMethod.cs
@@ -1,0 +1,18 @@
+using System.Runtime.Serialization;
+
+namespace W3C.Contracts.Matchmaking;
+
+public enum MatchSortMethod
+{
+    [EnumMember(Value = "startTimeDescending")]
+    StartTimeDescending,
+
+    [EnumMember(Value = "startTimeAscending")]
+    StartTimeAscending, // Currently not used, but adding it for future use
+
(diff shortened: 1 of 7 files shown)


MichaelDeitz:
@marcoabreu is this completed, or do we need to do more work? I see a few commits, but not sure what the state is.

PR #438, opened 2025-06-25
Do not log 'League was not found' twice to file

gustav87:


W3ChampionsStatisticService/PlayerProfiles/GameModeStats/GameModeStatQueryHandler.cs
```diff
@@ -76,8 +75,7 @@ private void PopulateLeague(
         }
         catch (Exception e)
         {
-            _trackingService.TrackException(e, $"A League was not found for {rank.Id} RankNumber: {rank.RankNumber} Leage: {rank.League}");
-            Log.Error($"A League was not found for {rank.Id} RankNumber: {rank.RankNumber} League: {rank.League} {e.Message}");
+            _trackingService.TrackException(e, $"A League was not found for {rank.Id} - RankNumber: {rank.RankNumber} - League: {rank.League} - Message: {e.Message}");
```
marcoabreu, line 78: Please keep the Log.error since that goes to the logging service

W3ChampionsStatisticService/Services/TrackingService.cs
```diff
@@ -45,7 +41,10 @@ public void TrackUnauthorizedRequest(string authorization, ControllerBase contro
 
     public void TrackException(Exception ex, string message)
     {
-        _logger.LogError(ex, message);
```
marcoabreu, line 48: We need the logger to log to grafana, don't we?

gustav87: This isn't serilog either

PR #440, opened 2025-06-29
Adds Block Player Functionality and Tests

francislavoie:
Implements the ability for users to block other players.

- Adds a `BlockPlayer` method to the `WebsiteBackendHub` to allow users to block other players.
- Implements various tests for the `WebsiteBackendHub` to ensure proper functionality and prevent regressions.
- Introduces interfaces for FriendRequestCache and FriendCommandHandler to make it possible to test.

Used Roo Code to make all these changes. I did run the tests locally, I installed .NET SDK 9.0 (latest) but it didn't work so I had to install 8.0 as well. Should be fine I guess. Very simple addition, just lots of tests to cover all the hub tasks.

### DIFF
--- a/W3ChampionsStatisticService/Hubs/WebsiteBackendHub.cs
+++ b/W3ChampionsStatisticService/Hubs/WebsiteBackendHub.cs
@@ -22,9 +22,9 @@ public class WebsiteBackendHub(
     IW3CAuthenticationService authenticationService,
     ConnectionMapping connections,
     IHttpContextAccessor contextAccessor,
-    FriendRequestCache friendRequestCache,
+    IFriendRequestCache friendRequestCache,
     IPersonalSettingsRepository personalSettingsRepository,
-    FriendCommandHandler friendCommandHandler,
+    IFriendCommandHandler friendCommandHandler,
     TracingService tracingService
 ) : Hub
 {
@@ -49,9 +49,9 @@ public class WebsiteBackendHub(
     private readonly IW3CAuthenticationService _authenticationService = authenticationService;
     private readonly ConnectionMapping _connections = connections;
     private readonly IHttpContextAccessor _contextAccessor = contextAccessor;
-    private readonly FriendRequestCache _friendRequestCache = friendRequestCache;
+    private readonly IFriendRequestCache _friendRequestCache = friendRequestCache;
     private readonly IPersonalSettingsRepository _personalSettingsRepository = personalSettingsRepository;
-    private readonly FriendCommandHandler _friendCommandHandler = friendCommandHandler;
+    private readonly IFriendCommandHandler _friendCommandHandler = friendCommandHandler;
     private readonly TracingService _tracingService = tracingService;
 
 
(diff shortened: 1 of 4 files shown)


W3ChampionsStatisticService/Hubs/WebsiteBackendHub.cs
```diff
@@ -276,6 +276,33 @@ await Clients.Caller.SendAsync(
...
+    public async Task BlockPlayer(string battleTag)
+    {
+        var currentUser = _connections.GetUser(Context.ConnectionId)?.BattleTag;
+        if (currentUser == null)
+        {
+            return;
+        }
+        try
+        {
+            var friendList = await _friendCommandHandler.LoadFriendList(currentUser);
```
marcoabreu, line 288: Could the CanBlock be a function of the friend thingy so we don't have to load the entire friendlist?

Basically delegating the responsibility of fetching the necessary data to the function that evaluates the block, rather than the caller?

W3ChampionsStatisticService/Hubs/WebsiteBackendHub.cs
```diff
@@ -276,6 +276,33 @@ await Clients.Caller.SendAsync(
...
+            CanBlock(friendList, battleTag);
+            friendList.BlockedBattleTags.Add(battleTag);
```
marcoabreu, line 290: Note that this here is teh WebsiteBackendHub, speak an API Handler. You'd generally want function in this file to not have business logic but only call downstream handlers that actually do the logic. So the adding to the blocklist etc should all be the downstream handlers and not this API handlers.

W3ChampionsStatisticService/Hubs/WebsiteBackendHub.cs
```diff
@@ -276,6 +276,33 @@ await Clients.Caller.SendAsync(
...
+                FriendResponseType.FriendResponseData.ToString(),
+                friendList,
+                null,
+                null,
+                $"Player {battleTag} blocked!"
+            );
+        }
+        catch (Exception ex)
+        {
+            await Clients.Caller.SendAsync(FriendResponseType.FriendResponseMessage.ToString(), ex.Message);
```
marcoabreu, line 302: Do we need to log this exception?

francislavoie: Not questions I feel equipped to answer tbh, it's the same code pattern as `UnblockFriendRequestsFromPlayer` directly below this fn 🤷‍♂️

WC3ChampionsStatisticService.UnitTests/Hubs/WebsiteBackendHubTests.cs
```diff
@@ -0,0 +1,567 @@
...
+
+        var friendList = new Friendlist("User#1234");
+        var friendListCache = new TestFriendListCache();
+        friendListCache.Upsert(friendList);
+        var friendRepository = new TestFriendRepository();
+        IFriendCommandHandler friendCommandHandler = new TestFriendCommandHandler(friendRepository, friendListCache);
+
+        var tracingService = new TracingService(new System.Diagnostics.ActivitySource("test"), new Mock<IHttpContextAccessor>().Object);
+
+        var hub = new WebsiteBackendHub(
```
marcoabreu, line 265: There is a lot of repeated boilerplate. Could this either be a parameterized test or part of a helper or the BeforeTest?

W3ChampionsStatisticService/Hubs/WebsiteBackendHub.cs
francislavoie: I'm just matching the existing patterns, I have no interest in refactoring this for the scope of this change.

WC3ChampionsStatisticService.UnitTests/Hubs/WebsiteBackendHubTests.cs
francislavoie: Reduced boilerplate, `+124 −327`

W3ChampionsStatisticService/Hubs/WebsiteBackendHub.cs
marcoabreu: I'd recommend you add a `using Serilog` at the top and then you do `Log.Error` (or `Log.Exception` if that's a thing) before you do the await.

W3ChampionsStatisticService/Hubs/WebsiteBackendHub.cs
francislavoie: Sure, but seems like a refactor out of scope of this

PR #445, opened 2025-07-12
re-enable friend list profile pictures

gustav87:


francislavoie:
Nice! Was there a change in the launcher to make the load lighter, in the end? I don't remember seeing one. You did do a lot with gzip etc so that might help in terms of bandwidth I guess idk

gustav87:
The launcher might reconnect less aggressively since then. I don't remember what's been done since this was disabled, but I disabled the pictures not because I specifically saw that it caused issues, but as a safety measure hoping to decrease load. If I remember correctly, website-backend struggled getting back up after a deploy.

marcoabreu:
Yeah we have exponential backoffs now

PR #447, opened 2025-07-12
use random default profile picture for `clan-and-picture` call if personal settings for player is missing

gustav87:


gustav87:
Chat service uses the `clan-and-picture` endpoint, so this avoids flickering default pics in chat. Creating personal settings if non-existent opens up creating personal settings for non-existent users. That could be mitigated, but perhaps this will do for now.

francislavoie:
```csharp
using Standart.Hash.xxHash; // Required for XXHash functionality
using System;
using System.Collections.Generic;
using System.Text; // Required for Encoding.UTF8

...

    public static int GetProfilePictureIndex(string battletag)
    {
        if (string.IsNullOrWhiteSpace(battletag))
        {
            // Handle empty or null battletags gracefully.
            // You might return a default index, throw an exception, or log an error.
            Console.WriteLine("Warning: Battletag cannot be null or empty. Returning index 0.");
            return 0;
        }

        // Convert the battletag string to a byte array using UTF8 encoding.
        // It's important to use a consistent encoding for hashing.
        byte[] battletagBytes = Encoding.UTF8.GetBytes(battletag);

        // Calculate the 64-bit XXHash of the battletag bytes.
        // A fixed seed (e.g., 0) ensures that the same input always produces the same hash.
        ulong hashValue = XXHash64.ComputeHash(battletagBytes, 0);

        // Use the modulo operator to get an index within the bounds of our list.
        // The result will be between 0 and DefaultProfilePictures.Count - 1.
        int index = (int)(hashValue % (ulong)DefaultProfilePictures.Count);

        return index;
    }
```

francislavoie:
> I don't think always getting the same random starter picture when you don't have a personal settings yet is something we need to have.

To be clear, it's still essentially random, just consistently random, using the battletag as a seed. It would just help avoid situations where refreshes to use different icons, and avoids actual randomness which makes writing tests harder, etc.

Issue #448, opened 2025-07-13
Players with duplicate friend requests get stuck


francislavoie:
<img width="1237" height="571" alt="Image" src="https://github.com/user-attachments/assets/b11eb7c8-a37e-47f1-b345-d2a3c91a5f7b" />

Reported on discord: https://discord.com/channels/688377338576109643/716611038950457393/1393758768994713621

Seems like the problem is here:

https://github.com/w3champions/website-backend/blob/2232aeeac0cae455858d93d85febd21b8c052193/W3ChampionsStatisticService/Friends/FriendRequestCache.cs#L49

Using `SingleOrDefault()` is the cause because it only returns successfully if there's zero or one results and failing if there's more. Using `FirstOrDefault()` could allow recovering more gracefully. See https://stackoverflow.com/a/1256784/846934

CoffMakesThings:
Should be fixed by https://github.com/w3champions/website-backend/pull/495

PR #453, opened 2025-08-04
Fix friend request cache duplicate bug (#448)

SifouByte:
Use FirstOrDefault instead of SingleOrDefault to avoid exceptions when duplicates exist, and remove any duplicate friend requests after returning the first found entry. See #448  for details.


CoffMakesThings:
Closing as it overlaps with https://github.com/w3champions/website-backend/pull/495

PR #465, opened 2025-09-08
Immutable player bans

gustav87:
website PR: https://github.com/w3champions/website/pull/923
matchmaking PR: https://github.com/w3champions/matchmaking-service/pull/684


marcoabreu:
That is not sufficient - a new ban will just override an existing ban. You need to change the entire datamodel and you still need to have deletes in the form of date changes.

PR #505, opened 2026-04-08
Add match duration filtering to /api/matches (Issue #472)

### LINKED ISSUE #472: Add match duration filter to search of finished games
Extend the search of finished games  `https://website-backend.w3champions.com/api/matches` with the following properties:

- Match duration length: Specify a match duration range (e.g. 300s - 99999s) which will filter to only return matches with a duration within that range.

This issue should include the backend work as well as the frontend work.

MichaelDeitz:
This PR implements the backend portion of Issue #472 by adding optional match duration filtering to the /api/matches endpoint.

Changes:
- Added minDuration and maxDuration parameters to MatchesController.GetMatches
- Updated MatchRepository.Load to apply duration filters
- Updated MatchRepository.Count to include duration filters
- Updated GetLoadFilter to support duration range
- Updated unit tests to match new method signatures

Notes:
- Backend implementation is complete and tested.
- Frontend changes are required in the website repo to expose the new filters.

Closes #472
Would need a new item created for the front end. If granted access can try to finish.

### DIFF
--- a/W3ChampionsStatisticService/Program.cs
+++ b/W3ChampionsStatisticService/Program.cs
@@ -107,7 +107,7 @@ builder.Services.AddSwaggerGen(f =>
 });
 
 // Configure and add MongoDB
-string mongoConnectionString = Environment.GetEnvironmentVariable("MONGO_CONNECTION_STRING") ?? "mongodb://157.90.1.251:3513"; // "mongodb://localhost:27017";
+var mongoConnectionString = Environment.GetEnvironmentVariable("MONGO_CONNECTION_STRING") ?? "mongodb://localhost:27017/w3champions";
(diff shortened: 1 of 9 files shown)


W3ChampionsStatisticService/Program.cs
```diff
@@ -107,7 +107,7 @@
 });
 
 // Configure and add MongoDB
-string mongoConnectionString = Environment.GetEnvironmentVariable("MONGO_CONNECTION_STRING") ?? "mongodb://157.90.1.251:3513"; // "mongodb://localhost:27017";
```
CoffMakesThings: This looks unintentionally left over from local testing.

PR #509, opened 2026-04-10
Fix: duration filter logic

MichaelDeitz:


W3ChampionsStatisticService/Matches/MatchRepository.cs
```diff
@@ -278,9 +278,25 @@ private FilterDefinition<Matchup> GetLoadFilter(int season, GameMode gameMode, H
                 && player.CurrentMmr <= maxMmr)));
         }
 
+        // Duration filter (ALWAYS applied)
```
CoffMakesThings: Really good. This PR is in a good state now. but please remove this comment before you merge. It is confusing because it implies that the duration filter is somehow ALWAYS applied in a way that others aren't.

CoffMakesThings (approved):
image

It works. Good job!

MichaelDeitz:
Do you want me to create a user story or enhancement request for the duration filter on the frontend?
Also, can you show where you would like it on the frontend?

CoffMakesThings:
> where you would like it on the frontend?

image

I would expect it here somewhere.

> Do you want me to create a user story or enhancement request for the duration filter on the frontend?

You can make an issue on the website repo if you want.

PR #511, opened 2026-04-11
Fix/now playing ongoing match lookup

### LINKED ISSUE #348: Matches page queries aren't being handled correctly

MichaelDeitz:
Fix: Now Playing ongoing match lookup. Should fix #348.

### DIFF
--- a/W3ChampionsStatisticService/Matches/MatchRepository.cs
+++ b/W3ChampionsStatisticService/Matches/MatchRepository.cs
@@ -294,7 +294,7 @@ public class MatchRepository(MongoClient mongoClient, IOngoingMatchesCache cache
         }
 
         return filter;
-        }
+    }
 
 
     public async Task InsertOnGoingMatch(OnGoingMatchup matchup)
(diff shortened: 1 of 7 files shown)


gustav87:
Does this PR contain the same commits you already merged in https://github.com/w3champions/website-backend/pull/509?

MichaelDeitz:
> Does this PR contain the same commits you already merged in #509?

I canceled this Pr as it was an incorrect one. #512 should fix it. I also added more information in #348.

PR #512, opened 2026-04-12
Add map filter to fix finished matches.

MichaelDeitz:
This fix updates the finished match map filtering for #348. I can now filter all matches correctly as long as the map exists in my database.
The map name used in the User Story did not exist in my DB. The actual stored value for Concealed Hill in my environment is:
Map: "s13_1ConcealedHill"
Because the database stores the raw map ID and not the readable map name, filtering must be done using the map field, not mapName.
This PR corrects the filtering logic, so finished matches can be filtered properly by map.


CoffMakesThings:
Does this change anything for you when you run website and website-backend locally? For me it doesn't.

PR #513, opened 2026-04-28
Fix map filter for finished matches

CoffMakesThings:
This fixes the map filter being broken for finished matches.

For seasons 10+, we have the MapName in the MatchUp collection.

In this PR we get the distinct MapNames for the season, offer them as dropdown options, and filter by them.

For seasons 1-10, we will stop offering the Map filter for now. Steps could be taken to make those work too, in a future PR.

This would benefit from an index on MatchUp.MapName.

image

[website pr](https://github.com/w3champions/website/pull/1040)

### DIFF
--- a/W3ChampionsStatisticService/Matches/MatchRepository.cs
+++ b/W3ChampionsStatisticService/Matches/MatchRepository.cs
@@ -233,14 +233,29 @@ public class MatchRepository(MongoClient mongoClient, IOngoingMatchesCache cache
             playerBlizzard.teamIndex);
     }
 
-    public async Task<List<Matchup>> Load(int season, GameMode gameMode, int offset = 0, int pageSize = 100, HeroType hero = HeroType.AllFilter, int minMmr = 0, int? maxMmr = null, int? minDuration = null, int? maxDuration = null)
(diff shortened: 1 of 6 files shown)


CoffMakesThings:
Been working on this with @MichaelDeitz

gustav87:
Can you use https://website-backend.w3champions.com/api/ladder/active-modes for fetching the maps?

gustav87:
Oh I see you're querying the maps played for the season from Matchup? If that is expensive to do for every request you might want to consider creating a new collection for it.

CoffMakesThings:
> Can you use https://website-backend.w3champions.com/api/ladder/active-modes for fetching the maps?

Isn't that only for the current season?

CoffMakesThings:
> Oh I see you're querying the maps played for the season from Matchup? If that is expensive to do for every request you might want to consider creating a new collection for it.

You mean the list of map names for the season? Yes. I think it is not so expensive if MapName is indexed, but maybe still a bit. What if we cache it on the BE (as well as the FE) for 10 minutes? Would that be good enough? As you said on the FE PR, they won't change much.

I'm a bit hesitant to create a new collection. Sounds complex.

gustav87:
Yeah that should do the trick!

gustav87:
IMO the frontend doesn't need a **timed** cache. If it has fetched the maps once, then it can assume those are valid. The user can just reload the page to clear it. But with a backend cache that's less important.

CoffMakesThings:
Ok! Will add, for this, timed caching to the BE and non-timed caching to the FE.

CoffMakesThings:
@gustav87 I have now added the 10 minute caching of map names on the BE

gustav87 (approved):

PR #514, opened 2026-04-29
randomracewithrace

MichaelDeitz:
Include Random race results in the race filter while preserving a separate Random match view; opponent race filtering is also supported.

### DIFF
--- a/W3ChampionsStatisticService/Matches/MatchRepository.cs
+++ b/W3ChampionsStatisticService/Matches/MatchRepository.cs
@@ -156,8 +156,8 @@ public class MatchRepository(MongoClient mongoClient, IOngoingMatchesCache cache
         }
         filter &= builder.Where(m => gameMode == GameMode.Undefined || m.GameMode == gameMode);
         filter &= builder.Where(m => gateWay == GateWay.Undefined || m.GateWay == gateWay);
-        filter &= builder.Where(m => playerRace == Race.Total || m.Teams.Any(team => team.Players[0].Race == playerRace && playerId == team.Players[0].BattleTag));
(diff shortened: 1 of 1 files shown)


francislavoie:
Hmm, what's the practical difference between looking at `[0]` and doing `.Any()`? Do we ever have more than one entry in `team.Players`?

MichaelDeitz:
Yes, team.Players can contain multiple entries since we support 1v1 through 4v4. [0] assumes a fixed ordering where the queried player is always first, but that is not guaranteed. In the repository, players are consistently identified by BattleTag, not position, and there is no enforced ordering in team.Players, so the current ordering looks like a result of presentation or query behavior. .Any() ensures we always match the correct player regardless of position. If ordering is ever guaranteed in the data model, we could simplify back to [0], but otherwise .Any() is the safe and correct approach. Happy to align with your approach, though, and change it.

francislavoie (approved):

MichaelDeitz:
I am not able to merge.  If you approve, you can merge and close. Thanks.

gustav87 (approved):

francislavoie:
FYI @MichaelDeitz I got some more feedback that some people like to use the search to see their MMR progress for a specific race, and now that we include random it messes with that a bit. I hadn't thought of it from that angle.

I think we should include a switch on the race picker dropdowns (screenshot for visual ref) just below the header ("Player Race") to toggle "With Random" (default off I guess). That would add an additional query param to the filter to say "Random>Orc alongside Orc" or just "Orc, not Random".

image

Example of what the switches look like (elsewhere)

image

PR #521, opened 2026-05-24
Store rating_lower_bound so it's available from game-mode-stats

francislavoie:
Currently, profiles in W3C launcher shows "Highest MMR", but only for "themselves" (when a player is looking at their own profile). This is because it's only being sent down to the launcher via the MM service instead of coming from the backend. The backend wasn't storing it when the MMR is updated. I think this is something useful we should store because it's an interesting stat. 

Of course we shouldn't show it when RD is still high (or I think the simpler condition is just "if rating_lower_bound is lower than the current MMR, just hide it", which has the same meaning effectively? I think? right...?)

I don't think this is really secret or "internal" info based on prior conversations, so I don't think there's much harm in doing this.


gustav87 (approved):

francislavoie:
Will be done differently with a formalizing of "highest MMR"

PR #528, opened 2026-06-08
P7-1: Lifetime win-milestone store

marcoabreu:
## Summary

Adds a permanent, downstream **lifetime win-milestone** read-model (P7-1) — a cosmetic "always something to climb" track, independent of rank, that never resets.

- New `ProgressionMilestone` read-model (collection `ProgressionMilestone`), keyed per entity (a player, or an arranged team) / game mode / race-for-`GM_1v1`, **season-less** so wins accumulate across all seasons. Mirrors the P8-2 `PlayerProgression` dedicated-repo triad (not the legacy `IPlayerRepository`).
- `ProgressionMilestoneHandler` ingests `MatchFinishedEvent`: increments `TotalWins` on the per-player `won` flag — AT players grouped by `atTeamId` into one shared team document, solo players individually — and records a compact rolling weekly activity window for every game (won or lost). Skips fake events.
- Pure `MilestoneTargetCalculator`: the next-milestone target is the next round-number of wins on a curve that coarsens as totals grow; a returning/low-activity player is given a nearer milestone (derived from their own recent activity), never a farther one. Computed **on read** — no stored target, no scheduled job.
- Dedicated `IProgressionMilestoneRepository` + impl; both repo and handler registered in `Program.cs` alongside the P8-2 read-model.
- New section in `docs/ranking/README.md` describing the read-model.

Idempotency is cursor-only at-least-once, consistent with the existing counting read-models (`PlayerOverallStats`, `GamesPerDay`).

## Test plan

- [x] `dotnet build` — 0 errors
- [x] `dotnet test` — 634 passed / 0 failed / 9 skipped (35 new tests across model, calculator, repository, handler — Mongo2Go + pure unit)
- [x] `dotnet format --verify-no-changes` — clean

## Checklist

- [x] Targets `prj/new-ranking-system` (not master/main)
- [x] Task: P7-1
- [x] Tests added (TDD) — Mongo2Go for store/handler/repo + pure unit tests for the calculator
- [x] Gate output green (build + test + format)
- [x] Additive, read-only downstream state; never feeds rank/MMR; no new feature flag (standard read-model handler, like the P8-2 ingest)
- [x] Per-repo doc updated (`docs/ranking/README.md`)
- [x] Public-repo boundary respected (no internal mechanism or roadmap detail; only the player-facing win/milestone contract)

### DIFF
--- a/W3ChampionsStatisticService/PlayerProfiles/ProgressionStats/MilestoneTargetCalculator.cs
+++ b/W3ChampionsStatisticService/PlayerProfiles/ProgressionStats/MilestoneTargetCalculator.cs
@@ -1,19 +1,25 @@
+using System;
+
 namespace W3ChampionsStatisticService.PlayerProfiles.ProgressionStats;
 
 // Pure, on-read next-milestone selection. Baseline granularity curve (the next round-number of
-// wins; the step coarsens as totals grow) plus a player-local catch-up that only ever NARROWS the
-// gap (a returning/low-volume player gets a nearer milestone; an active player gets the baseline —
-// never a widening gap). All thresholds are tunable constants; tests assert behaviour + invariants.
+// wins; the step coarsens as totals grow, capped at 100) plus a player-local catch-up that only
+// ever NARROWS the gap (a returning/low-volume player gets a nearer milestone; an active player
+// gets the baseline — never a widening gap). All thresholds are tunable constants; tests assert
+// behaviour + invariants.
 public static class MilestoneTargetCalculator
 {
-    // (upperBoundExclusive, step). Coarser steps are integer multiples of finer ones so a finer
-    // catch-up step never produces a farther target.
+    // (upperBoundExclusive, step). Step sizes are 5, 10, 25, 50, 100 — capped at 100. They are NOT
+    // all integer multiples of one another (25 is not a multiple of 10), so a finer catch-up step can
+    // occasionally land beyond the baseline milestone; Compute clamps the catch-up to the baseline so
+    // it can never widen the gap regardless of the step set.
     private static readonly (long UpperExclusive, long Step)[] Bands =
     {
         (50, 5),
+        (100, 10),
+        (250, 25),
         (500, 50),
-        (5000, 250),
-        (long.MaxValue, 1000),
+        (long.MaxValue, 100),
     };
 
     // Catch-up thresholds over the trailing recent-activity window (tunable).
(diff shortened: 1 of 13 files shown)


W3ChampionsStatisticService/PlayerProfiles/ProgressionStats/MilestoneTargetCalculator.cs
```diff
@@ -0,0 +1,60 @@
...
+// Pure, on-read next-milestone selection. Baseline granularity curve (the next round-number of
+// wins; the step coarsens as totals grow) plus a player-local catch-up that only ever NARROWS the
+// gap (a returning/low-volume player gets a nearer milestone; an active player gets the baseline —
+// never a widening gap). All thresholds are tunable constants; tests assert behaviour + invariants.
+public static class MilestoneTargetCalculator
+{
+    // (upperBoundExclusive, step). Coarser steps are integer multiples of finer ones so a finer
+    // catch-up step never produces a farther target.
+    private static readonly (long UpperExclusive, long Step)[] Bands =
+    {
```
marcoabreu, line 17: I think the maximum Step should be 100. Other steps are 5, 10, 25, 50, 100.

W3ChampionsStatisticService/PlayerProfiles/ProgressionStats/ProgressionMilestoneHandler.cs
```diff
@@ -0,0 +1,78 @@
...
+
+    private async Task RecordEntity(Match match, List<PlayerMMrChange> entityPlayers, DateTimeOffset playedAt)
+    {
+        var first = entityPlayers[0];
+        var playerIds = entityPlayers.Select(p => PlayerId.Create(p.battleTag)).ToList();
+        var gameMode = GameModesHelper.ToArrangedTeamVariant(match.gameMode, first.IsAt);
+        // Milestones accumulate across seasons, so the key must be season-less and stable.
+        // Unlike the season-keyed ladder (which gates race on RaceSplitStartSeason), we
+        // always include race for solo GM_1v1 so pre-season-2 wins roll up into the same doc.
+        var race = gameMode == GameMode.GM_1v1
```
marcoabreu: Re-use the helper that we have regarding the "gamemode supports race"

W3ChampionsStatisticService/PlayerProfiles/ProgressionStats/MilestoneTargetCalculator.cs
marcoabreu: Done in 1586cd6 — capped the max step at 100; the step set is now 5/10/25/50/100 (bands: <50→5, <100→10, <250→25, <500→50, then 100). One subtlety: 25 isn't an integer multiple of 10, which would break the old "finer catch-up step never widens" guarantee, so the catch-up now clamps to the baseline target (`min`) — it can only ever narrow, for any step set. Tests updated, incl. an explicit max-step-is-100 case.

W3ChampionsStatisticService/PlayerProfiles/ProgressionStats/ProgressionMilestoneHandler.cs
marcoabreu: Done in 1586cd6 — added `GameModesHelper.IsRaceSplitGameMode(gameMode)` (matching the existing `IsFfaGameMode`/`IsMeleeGameMode` convention) and the handler now calls it instead of inlining `== GM_1v1`. `UsesRaceInLadderKey` now delegates to it (`IsRaceSplitGameMode(gameMode) && season >= RaceSplitStartSeason`). The milestone store intentionally uses the season-independent helper (no season gate), since its key is season-less/cross-season.

PR #533, opened 2026-06-10
feat(progression): apex leaderboard + progression league ladder endpoints (P11-4)

marcoabreu:
Adds two public ladder surfaces for the progression ranking system: the **apex leaderboard** (Grand Master + Master standings) and the **non-apex progression league ladder** (Adept…Grass). Downstream half of P11-4.

## Apex leaderboard — `GET /api/ladder/apex?season=&gameMode=`
The daily Grand Master + Master cohort, ordered by apex points (GM first), with the floating GM cutoff. It is kept current by ingesting a new `ApexStandingsChangedEvent` through the same ISyncable checkout path the ladder already uses (new `ApexSyncHandler` → `ApexLeaderboard` read-model), then served by an anonymous endpoint. Served from the synced read-model rather than the per-player progression collection, because the daily Grand Master/Master assignment changes without a match.
- Response: `{ cutoffApexPoints, gmCount, players: [{ playersInfo, apexPoints, league, rankNumber }] }` (`league` 0 = Grand Master, 1 = Master). Gateway-agnostic.

## Progression league ladder — `GET /api/ladder/progression?season=&gameMode=&league=&division=&race=`
Non-apex league/division pages served directly from the existing `PlayerProgression` read-model (always current for these leagues), backed by a new compound index. Returns the **same `Rank` shape** as `/api/ladder/{leagueId}` so the existing ladder grid renders it unchanged. Points-desc with a global `rankNumber`; `take` capped at 500; apex leagues (0/1) return empty (use `/api/ladder/apex`).

## Notes
- Enrichment DRY: the per-player display enrichment was extracted from `PopulatePlayerInfos` into shared helpers reused by the existing ladder, apex, and progression paths (behavior-preserving — existing `RankTests` green).
- New compound index on `PlayerProgression` `(Season, GameMode, League, Division, Race, Points desc)` via the standard `IRequiresIndexes` startup path.

## Checklist
- [x] Targets `prj/new-ranking-system`
- [x] Task: P11-4 (Workstreams B+C — website-backend)
- [x] Tests (TDD, NUnit + Mongo2Go): apex ingest/handler/repo/endpoint (18) + progression index/query/ladder (incl. apex-guard both leagues, paging, race split, empty cases); full suite 719 passed / 0 failed
- [x] Gates green: `dotnet build` 0 · `dotnet test` · `dotnet format --verify-no-changes` exit 0
- [x] Public-repo boundary: whole diff + commit messages scanned — only league/division/points/apexPoints/Grand Master/Master/cutoff vocabulary; no internal mechanism, no private-repo names
- [x] Anonymous, gateway-agnostic endpoints; additive (legacy ladder untouched)
- [x] Per-repo doc updated (`docs/ranking/README.md`)

### DIFF
--- a/W3ChampionsStatisticService/PlayerProfiles/ProgressionStats/PlayerProgressionRepository.cs
+++ b/W3ChampionsStatisticService/PlayerProfiles/ProgressionStats/PlayerProgressionRepository.cs
@@ -2,6 +2,8 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
 using MongoDB.Driver;
+using W3C.Contracts.GameObjects;
+using W3C.Contracts.Matchmaking;
 using W3C.Domain.Repositories;
 using W3C.Domain.Tracing;
 using W3ChampionsStatisticService.Ports;
@@ -10,8 +12,14 @@ namespace W3ChampionsStatisticService.PlayerProfiles.ProgressionStats;
 
 [Trace]
 public class PlayerProgressionRepository(MongoClient mongoClient)
-    : MongoDbRepositoryBase(mongoClient), IPlayerProgressionRepository
+    : MongoDbRepositoryBase(mongoClient), IPlayerProgressionRepository, IRequiresIndexes
 {
+    // Grand Master and Master are the apex leagues; they are served by the dedicated
+    // /api/ladder/apex endpoint, not the per-league progression ladder below.
+    private static bool IsApexLeague(int league) => league < (int)ProgressionLeague.Adept;
+
+    public string CollectionName => nameof(PlayerProgression);
+
     public Task<PlayerProgression> LoadProgression(string id)
     {
         return LoadFirst<PlayerProgression>(id);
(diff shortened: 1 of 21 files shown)


W3ChampionsStatisticService/PlayerProfiles/ProgressionStats/PlayerProgressionRepository.cs
```diff
@@ -10,8 +12,14 @@ namespace W3ChampionsStatisticService.PlayerProfiles.ProgressionStats;
 
 [Trace]
 public class PlayerProgressionRepository(MongoClient mongoClient)
-    : MongoDbRepositoryBase(mongoClient), IPlayerProgressionRepository
+    : MongoDbRepositoryBase(mongoClient), IPlayerProgressionRepository, IRequiresIndexes
 {
+    // Leagues 0 (Grand Master) and 1 (Master) are the apex leagues; they are served by the
+    // dedicated /api/ladder/apex endpoint, not the per-league progression ladder below.
+    private const int FirstNonApexLeague = 2;
```
marcoabreu: Check if we have a central place where we define progression-related vairables. Specifically, if we already have an enum for leagues so that we don't ahve to reside to raw numbers

marcoabreu: There was no progression-league enum in the repo — `PrestigeRankComparer` documented the encoding in a comment and compared raw `.League` ints. Added `ProgressionLeague` (GrandMaster=0 … Grass=8) in `PlayerProfiles/ProgressionStats` with the authoritative-encoding XML doc. The apex guard is now `IsApexLeague(league) => league < (int)ProgressionLeague.Adept` and the `/api/ladder/progression` default is `(int)ProgressionLeague.Adept` (no more raw `2`). The passthrough `league` fields (sync handler / query mapping) carry the int through with no literal. Left `PrestigeRankComparer` as-is: it's relational-only (no literals), and its `PeakRank.League` is a serialized `int?` whose conversion would ripple through the prestige read-model — out of scope here.

PR #534, opened 2026-06-11
feat(progression): bracket population metrics on /metrics (P10-2)

marcoabreu:
## P10-2 — website-backend bracket-population metrics

Adds opt-in Prometheus gauges describing the progression ladder's bracket populations, published on the existing authenticated `/metrics` endpoint (scraped by the `websitebackend` job).

### What
- `CountByBracket(season)` + `LoadMaxSeason()` aggregation queries over the `PlayerProgression` read-model (server-side `$match`/`$group`).
- `ProgressionBracketMetrics` — two gauges, refreshed wholesale each pass (stale series dropped on bracket-empty / season rollover):
  - `progression_bracket_count{gameMode,league,division}`
  - `progression_ranked_total{gameMode}`
- `ProgressionBracketMetricsService` — `BackgroundService`, 15-minute refresh, current season = highest season present.
- Registered only when `PROGRESSION_METRICS_ENABLED=true` (off by default → inert).

### Tests (TDD)
- Repository aggregation tests via Mongo2Go (bracket grouping, unplaced/other-season exclusion, max-season).
- Gauge tests via the prometheus-net registry (population, stale-series removal, division-less label), with explicit per-test isolation.

### Gates
- `dotnet build`: 0 errors
- `dotnet test` (excl. shared-remote-Mongo `ClanTests`/`IntegrationTestBase` flakes): 690 passed / 0 failed / 9 skipped
- `dotnet format --verify-no-changes`: clean

### Notes
- Inert until the env flag is set; emits only public ladder display vocabulary (league/division/points/season/gameMode).
- Dashboards + alert rules for these series live in the ops Prometheus/Grafana repo.

Targets `prj/new-ranking-system` (the long-lived feature branch), not `main`.

### DIFF
--- a/W3ChampionsStatisticService/PlayerProfiles/ProgressionStats/PlayerProgressionRepository.cs
+++ b/W3ChampionsStatisticService/PlayerProfiles/ProgressionStats/PlayerProgressionRepository.cs
@@ -88,6 +88,9 @@ public class PlayerProgressionRepository(MongoClient mongoClient)
     public async Task<int?> LoadMaxSeason()
     {
         var collection = CreateCollection<PlayerProgression>();
+        // SortByDescending(Season).Limit(1) is served as a cheap reverse scan of the existing
+        // Season-leading compound index. A future index refactor must keep Season as a leading
+        // (ascending) key or this degrades to a COLLSCAN + in-memory sort.
         var latest = await collection
             .Find(Builders<PlayerProgression>.Filter.Empty)
             .SortByDescending(p => p.Season)
--- a/W3ChampionsStatisticService/PlayerProfiles/ProgressionStats/PlayerProgressionRepository.cs
+++ b/W3ChampionsStatisticService/PlayerProfiles/ProgressionStats/PlayerProgressionRepository.cs
@@ -64,6 +64,38 @@ public class PlayerProgressionRepository(MongoClient mongoClient)
             .ToListAsync();
     }
 
+    public async Task<List<ProgressionBracketCount>> CountByBracket(int season)
+    {
+        var collection = CreateCollection<PlayerProgression>();
+        var filter = Builders<PlayerProgression>.Filter.Eq(p => p.Season, season)
+            & Builders<PlayerProgression>.Filter.Ne(p => p.League, null);
+        var grouped = await collection.Aggregate()
+            .Match(filter)
+            .Group(p => new { p.GameMode, p.League, p.Division },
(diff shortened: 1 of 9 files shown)


W3ChampionsStatisticService/PlayerProfiles/ProgressionStats/PlayerProgressionRepository.cs
```diff
@@ -64,6 +64,41 @@ public async Task<List<PlayerProgression>> LoadPlayersByProgressionLeague(
...
+            {
+                GameMode = g.Key.GameMode,
+                League = g.Key.League!.Value,
+                Division = g.Key.Division,
+                Count = g.Count,
+            })
+            .ToList();
+    }
+
+    public async Task<int?> LoadMaxSeason()
```
marcoabreu: The service already knows the active season, doesn't it?

W3ChampionsStatisticService/Program.cs
```diff
@@ -247,6 +247,11 @@
 builder.Services.AddHostedService<W3ChampionsStatisticService.Common.Services.MongoIndexInitializationService>();
 builder.Services.AddHostedService<W3ChampionsStatisticService.Services.BackgroundTasks.UpdateMaxMmrService>();
 
+if (Environment.GetEnvironmentVariable("PROGRESSION_METRICS_ENABLED") == "true")
```
marcoabreu: We dont use env-vars for the progression system. Just have this on

W3ChampionsStatisticService/PlayerProfiles/ProgressionStats/PlayerProgressionRepository.cs
marcoabreu: Good call — switched to the canonical ladder season. `ProgressionBracketMetricsService` now resolves the current season from `IRankRepository.LoadSeasons()` (max `Season.Id`, the same source `LadderController` uses), instead of a bespoke max-query on the read-model. Removed `LoadMaxSeason()` from the interface, repository, and its test. Done in 0240d51.

W3ChampionsStatisticService/Program.cs
marcoabreu: Removed the `PROGRESSION_METRICS_ENABLED` env gate — the hosted service is now registered unconditionally, right after `UpdateMaxMmrService` (always on). Done in 0240d51.

PR #535, opened 2026-06-17
feat: tag lag reports with player-side network category (LAN / LastMile)

marcoabreu:
## Summary

Adds optional **`tags`** to lag-report submissions so player-side network issues can be categorised and filtered. The desktop client classifies whether a post-game connection problem was on the player's local network (`LAN`) or the link to their ISP (`LastMile`) and attaches the corresponding tag to the background lag report it already submits.

This PR is the backend half: accept, persist, index, filter, and surface those tags.

## Changes

- `ELagReportTag` enum (`LAN`, `LastMile`) — member names are the wire strings via the existing `JsonStringEnumListConverter`.
- `tags` on the submission DTO — tolerant of the field being **absent** (older clients) and of unknown values.
- Persist `Tags` on `LagReportPlayer` (BSON int array); map from the submission DTO.
- Index `Players.Tags`; add a case-insensitive `tag` filter to the report query endpoint (unknown tag ⇒ no filter, not an error).
- Surface `Tags` in the admin list projection.
- Cap the number of tags in submission validation (mirrors the existing categories cap).
- Boyscout: the validation method's inline numeric caps are now named constants.

## Compatibility / rollout

Fully backward-compatible — `tags` absent behaves exactly as empty, so this can (and should) **deploy before** the client change ships. No migration required.

## Tests

Isolated Mongo2Go coverage: BSON round-trip, multi-player single-game merge, filter-by-tag (positive + negative), case-insensitivity, index existence, list-projection retention, and the validation cap boundary. `dotnet format --verify-no-changes` clean.

## Test plan

- [ ] CI green
- [ ] Submit a report with `tags: ["LAN"]` ⇒ persists and is filterable; submit one with no `tags` ⇒ no regression

### DIFF
--- a/W3ChampionsStatisticService/LagReports/LagReport.cs
+++ b/W3ChampionsStatisticService/LagReports/LagReport.cs
@@ -98,6 +98,12 @@ public class LagReportPlayer
     [JsonConverter(typeof(JsonStringEnumListConverter<EIssueCategory>))]
     public List<EIssueCategory> IssueCategories { get; set; } = [];
 
+    /// <summary>
+    /// System-derived connection-issue verdict tags from the launcher
+    /// (e.g. LAN, LastMile). Distinct from the player-selected IssueCategories.
+    /// </summary>
+    public List<ELagReportTag> Tags { get; set; } = [];
+
     /// <summary>Free-text description from the player.</summary>
     public string FreeText { get; set; } = "";
 
--- a/W3ChampionsStatisticService/LagReports/LagReport.cs
+++ b/W3ChampionsStatisticService/LagReports/LagReport.cs
@@ -102,7 +102,7 @@ public class LagReportPlayer
     /// System-derived connection-issue verdict tags from the launcher
     /// (e.g. LAN, LastMile). Distinct from the player-selected IssueCategories.
     /// </summary>
-    public List<ELagReportTag> Tags { get; set; } = [];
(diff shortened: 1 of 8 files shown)


W3ChampionsStatisticService/LagReports/LagReport.cs
```diff
@@ -98,6 +98,12 @@ public class LagReportPlayer
     [JsonConverter(typeof(JsonStringEnumListConverter<EIssueCategory>))]
     public List<EIssueCategory> IssueCategories { get; set; } = [];
 
+    /// <summary>
+    /// System-derived connection-issue verdict tags from the launcher
+    /// (e.g. LAN, LastMile). Distinct from the player-selected IssueCategories.
+    /// </summary>
+    public List<ELagReportTag> Tags { get; set; } = [];
```
marcoabreu: Tags is a quite generic name; should we think about a better name?

marcoabreu: Good call. Renamed to `ConnectionIssueTags` (acd6600), then made it consistent end-to-end in 7c80c8d:
- wire JSON field `connection_issue_tags` (snake_case, matching `free_text` / `lag_events`)
- BSON field + index `ConnectionIssueTags` / `Players.ConnectionIssueTags`
- query filter param `connection_issue_tag`
- launcher submission renamed in lockstep

The `ELagReportTag` enum (values `LAN` / `LastMile`) is unchanged.

PR #538, opened 2026-07-05
Chat revamp: website-backend integration (W1–W3)

marcoabreu:
## Chat revamp — website-backend integration

website-backend's slice of the chat revamp: chat-profile enrichment, the moderation proxy re-point, friend-presence retirement, block/friend change-pings, and the relationship-snapshot endpoint chat reads. Design approved 2026-07-03; spec/plans live in the separate `prj-chat-rewrite` workspace (not committed here).

### What's included (3 workstreams)
| Item | Scope |
|---|---|
| **W1** | Chat-profile endpoint — rank/games/season enrichment on `ChatDetailsDto` (additive, fail-soft; legacy wire shape byte-identical) |
| **W3** | Moderation proxy re-pointed to chat's new paged `/api/moderation/...`; friend-presence (`FriendOnlineStatus`) retirement — **cutover-gated** |
| **W2** | Outbound HMAC change-pings on block/friend mutations + inbound service-authed relationship-snapshot endpoint |

### Gate evidence (Tier-0 independently re-run)
`dotnet build` → 0 errors (1 pre-existing unrelated warning) · `dotnet format --verify-no-changes` → clean · `dotnet test` (unit) → **750 passed / 9 skipped / 0 failed**. (The `IntegrationTestBase` suite hits a shared **remote** Mongo and is excluded from the gate — none of W1–W3's tests depend on it.)

### Change-ping / relationship contract (matches chat-service C7 + C5)
- **Outbound** `RelationshipChangeNotifier` → `POST {chat}/internal/relationship-changes`, best-effort **fire-and-forget after commit** (a ping failure never surfaces to the hub caller or adds latency — proven by test). Signs `"v1."+ts+"."+rawBody` with headers `X-W3C-Webhook-Timestamp` + `X-W3C-Signature: v1=<hex>`.
- **Inbound** `GET /api/players/{battleTag}/chat-relationships` → `{ "friends": [], "blocked": [] }`, guarded by `ChatServiceSecretAuthFilter` (**fail-closed**: missing/wrong/unconfigured secret → 401).

### Decisions / notes for you
- **Presence retirement is gated, default OFF:** `RETIRE_FRIEND_ONLINE_STATUS` unset ⇒ `FriendOnlineStatus` still emits. **Flip to retire only in the coordinated cutover** — chat's `FriendPresenceChanged` (connect+disconnect) is the replacement; retiring early strips live launchers of friend presence. `ConnectionMapping` retained for friends-CRUD push.
- **W1 legacy JSON keys** `chatColor`/`chatIcons` kept (not §4's `selectedChatColor`/`selectedChatIcons`) to avoid breaking existing consumers — a wire-shape test pins them; confirm §4 was descriptive, not literal.
- W1 `LoadRanksForPlayers` does a full-season scan (~20–40 ms/connect) — indexed point-lookup is an ops-owned follow-up.
- `/api/loungeMute` proxying and tolerant `Enum.TryParse` `EPermission` handling are **unchanged** (verified untouched).

### I2 / deploy gates
- Env: `CHAT_INTERNAL_API_SECRET` (outbound signing; unset ⇒ pings cleanly disabled) and `CHAT_RELATIONSHIPS_API_SECRET` (inbound auth; unset ⇒ fail-closed 401). The latter must equal chat-service's `STATISTIC_SERVICE_RELATIONSHIPS_SECRET`. Distinct from mm's secret.
- Ship W2+W3 only in the coordinated cutover release (chat is rebuilt in place at the same URL; the old prod chat has no paged endpoint / no presence replacement until cutover).

Built via plan → per-task implement+review → whole-branch review (security-reviewer on the signing + snapshot-auth surfaces). Reviewed pre-merge; opening as **draft** for your review.


W3C.Domain/ChatService/ChatServiceClient.cs
```diff
@@ -82,9 +92,29 @@ public async Task<string> DeleteLoungeMute(string battleTag, [NoTrace] string au
...
+
+    [NoTrace]
+    public async Task<ModerationMessagePageDto> GetModerationChannelMessages(string channelId, string authorization)
     {
-        string url = $"{ChatServiceApiUrl}/api/chat/{chatRoom}";
+        // Defensive encode: channelId is expected to always be a server-resolved value (see
+        // GetChatRoomMessages), never raw caller input, but this belt-and-suspenders encode keeps
+        // the method safe against path-segment injection even if a future caller gets it wrong.
+        string encodedChannelId = HttpUtility.UrlEncode(channelId);
+        string url = $"{ChatServiceApiUrl}/api/moderation/channels/{encodedChannelId}/messages?limit=100";
```
marcoabreu: Is there a way to get messages beyond the initial 100?

W3C.Domain/ChatService/ChatServiceClient.cs
```diff
@@ -93,16 +123,89 @@ public async Task<ChatMessage[]> GetChatRoomMessages(string chatRoom, string aut
         {
             throw new HttpRequestException(content, null, response.StatusCode);
         }
-        var deserializeObject = JsonConvert.DeserializeObject<ChatMessageDto[]>(content);
+        var deserializeObject = JsonConvert.DeserializeObject<ModerationMessagePageDto>(content);
+        return deserializeObject;
+    }
+
+    /// <summary>
+    /// Legacy-parity compatibility shim: resolves the caller-supplied room NAME to a chat-service
```
marcoabreu: Why do we need a legacy-parity compatibility shim?

W3C.Domain/ChatService/ChatServiceClient.cs
```diff
@@ -93,16 +123,89 @@ public async Task<ChatMessage[]> GetChatRoomMessages(string chatRoom, string aut
...
+                page = await GetModerationChannelMessages(channelId, authorization);
+            }
+            catch (HttpRequestException retryEx) when (retryEx.StatusCode == HttpStatusCode.NotFound)
+            {
+                return [];
+            }
+        }
+
+        return [.. (page?.Messages ?? [])
+            .Where(m => !m.Deleted && !m.Shadow)
```
marcoabreu: Isn't this here an admin method? In that case, shouldn't we include and flag deleted and shadow messages so a moderator can see them?

W3ChampionsStatisticService/Hubs/PresenceSettings.cs
```diff
@@ -0,0 +1,28 @@
...
+/// Chat-service now owns presence via its own FriendPresenceChanged system; this flag
+/// exists purely so the I2 cutover release can retire the hub broadcast once every
+/// launcher consumer has migrated off it.
+///
+/// Driven by the <c>RETIRE_FRIEND_ONLINE_STATUS</c> environment variable: "true"
+/// (case-insensitive) retires the broadcast (stop emitting); absent or any other value
+/// keeps emitting, which is the current/default behavior and MUST stay the default
+/// until the I2 cutover ships.
+///
+/// Once the cutover has stabilized, this whole pathway (<c>NotifyFriendsWithIsOnline</c>,
```
marcoabreu: Do need for a setting-based cutover. We will immediately switch over as this service is deployed - no env var

W3ChampionsStatisticService/PlayerProfiles/PlayerRepository.cs
```diff
@@ -129,6 +129,13 @@ public Task<List<PlayerGameModeStatPerGateway>> LoadGameModeStatPerGateway(
             t.Season == season);
     }
 
+    public Task<List<PlayerGameModeStatPerGateway>> LoadGameModeStatPerGateway(string battleTag, int season)
+    {
+        return LoadAll<PlayerGameModeStatPerGateway>(t =>
```
marcoabreu, line 173: How efficient/expensive is this query, considering it will be querying a lot?

marcoabreu:
## Review feedback addressed — remediation pushed (`c97167b`)

All five points handled in 4 commits (`e94d93e..c97167b`). Build 0 errors, `format --verify-no-changes` clean, targeted suite **109 tests** green (unit + Mongo2Go repo/index integration; full `IntegrationTestBase` deliberately not run — it hits shared remote Mongo, a known flake, and nothing here depends on it).

**`ChatServiceClient.cs:117` — "is there a way to get messages beyond the initial 100?"** — Yes, now (`5f95279`). New `GetModerationChannelHistory(chatRoom, beforeSeq, limit, [NoTrace] auth)` plumbs the chat-service cursor (`beforeSeq` appended only when non-null; `limit` clamped `[1,100]`), exposed as `GET /api/moderation/launcher-chat/{chatRoom}/messages?beforeSeq=&limit=` → `{ messages[], nextBeforeSeq }`. Paging back through full history is a `nextBeforeSeq` walk.

**`ChatServiceClient.cs:131` — "why do we need a legacy-parity compatibility shim?"** — We don't anymore — retired it (`5f95279`). `GetChatRoomMessages` is now a thin filtering adapter over the shared method above (the resolve/404-evict-retry-once logic is deduped, no more copy). The legacy `GET .../launcher-chat/{chatRoom}` endpoint keeps its exact `{id,message,time,battleTag}[]` shape purely so the current website page keeps working. **Follow-up (website repo, not these PRs):** once `admin-launcher-chat` migrates to the new `/messages` endpoint, `GetChatRoomMessages` + `ChatMessage` become dead and can be deleted outright.

**`ChatServiceClient.cs:179` — "isn't this an admin method? shouldn't it include+flag deleted and shadow messages so a moderator can see them?"** — Agreed, and the new endpoint does exactly that (`5f95279`). `ModerationMessageDto` is enriched with `Seq`, `SenderName`, `DeletedBy`, `DeletedAt`, and the mapper returns **every** row including deleted/shadow (it does *not* filter) with the flags attached, so a moderator sees the full record. The legacy endpoint stays filtered for page-compat only.

**`PresenceSettings.cs:17` — "no need for a setting-based cutover; we switch over immediately on deploy — no env var"** — Done, fully removed (`c97167b`). Deleted `PresenceSettings.cs`, the env gate, the `Program.cs` registration, the hub ctor param, `NotifyFriendsWithIsOnline` + **both** call sites, and the `FriendResponseType.FriendOnlineStatus` member. `FriendOnlineStatus` is no longer emitted anywhere (two new guard tests assert `Times.Never` on connect + disconnect). `ConnectionMapping` and all friends-CRUD push (incl. the `FriendsWithPictures` `IsOnline` snapshot) are preserved — presence just moves to chat-service on deploy.

**`PlayerRepository.cs:134` — "how efficient/expensive is this query, given it runs a lot?"** — Checked against prod (`f0ae860`). The query `{PlayerIds.BattleTag, Season}` already rides an **IXSCAN**: prod `explain(executionStats)` picks `battleTag_gw_season`, **totalKeysExamined=3, totalDocsExamined=1, 0ms**. So it's cheap — the real latent risk was that both prod indexes were provisioned **out-of-band, not code-managed**, so a fresh/rebuilt env would COLLSCAN ~649k docs on every chat-user render. Fix: `PlayerRepository` now implements `IRequiresIndexes` declaring both indexes with exact prod spec (idempotent no-op against prod), ensured at startup by `MongoIndexInitializationService`. No query rewrite.

---
**One design decision to flag (ticket-mint hub auth, `de83c88`):** `WebsiteBackendHub` is consumed by **both** the launcher and the **browser website**. A browser can't hold a JWT in Rust, so a tickets-only hard cutover (like chat-service's launcher-only hub) would break production browser users on deploy. So `OnConnectedAsync` is **ticket-first, raw-JWT fallback** — the launcher stops exposing the JWT to JS (closed launcher-side by #833/LE-2), while browsers keep the existing JWT path. A 64-hex ticket never validates as a JWT and vice-versa, so the two inputs are unambiguous. This is a forced consequence of the shared hub; calling it out explicitly in case you'd prefer a different browser-auth story.

marcoabreu:
### Follow-up (`8337950`) — hard-cutover to ticket-only, dual-accept removed

You were right to challenge the dual-accept. Verified the ground truth: **the browser website does not connect to `WebsiteBackendHub` at all** — zero SignalR in the site since it was removed in #122; browsers are REST + `Authorization: Bearer <JWT>`. The **launcher is the hub's only client**, so WB-1's "browsers can't hold a JWT in Rust" rationale was false.

Per your call — hard-cutover to **ticket-only**: removed the raw-JWT fallback branch from `OnConnectedAsync` (and the now-dead `IW3CAuthenticationService` hub dependency, so the fallback is structurally gone), which also closes the `validateLifetime:false` expired-JWT gap that lived only on that path. Corrected the misleading "dual-accept / browser fallback" comments to the ticket-only rationale.

Deliberate rollout consequence you accepted: **not-yet-updated launchers lose wb-hub friends/presence on deploy** until they update — same forced-update window as the chat cutover. security-reviewer clean (independently grep-confirmed the website has no SignalR to this hub); 52 targeted tests pass.

PR #539, opened 2026-07-24
Add matches/search-opponents endpoint

DanyPell:
Backs the new match history opponent search on the website: w3champions/website#1103

- `GET api/matches/search-opponents?playerId=&search=&season=&gateWay=&gameMode=&limit=` — returns `[{ battleTag, matchCount, wins, losses }]`, sorted by games together; wins/losses are the searched player's record across those matches (allies share the result), scoped to the game-mode filter like matchCount
- one aggregation scoped to the player's own season matches: resolve the player's result per match, project battleTags + mode → unwind → optional contains-regex → group + count → top 10, so only ~10 tiny docs ever leave mongo
- allies in team games count as shared matches, same semantics as the existing `opponentId` filter — covered in the repo test incl. a 2v2
- results cached 60s; measured ~70–190ms cold on a seeded 4000-game / 600-opponent account, which is faster than the existing history page query on the same data
- `EnsureIndexesAsync` now creates `{ Teams.Players.BattleTag: 1, Season: -1 }` — this endpoint and the existing per-player history query both ride it

### DIFF
--- a/W3ChampionsStatisticService/Matches/MatchRepository.cs
+++ b/W3ChampionsStatisticService/Matches/MatchRepository.cs
@@ -5,6 +5,7 @@ using Serilog;
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 using W3C.Contracts.GameObjects;
 using W3C.Contracts.Matchmaking;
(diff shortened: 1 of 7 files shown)


W3ChampionsStatisticService/Matches/OpponentInfo.cs
```diff
@@ -0,0 +1,9 @@
+namespace W3ChampionsStatisticService.Matches;
+
+// One result of the player-scoped opponent search: a player who shares
+// finished matches with the searched player, and how many they share.
+public class OpponentInfo
+{
+    public string BattleTag { get; set; }
+    public long MatchCount { get; set; }
```
francislavoie, line 9: Should we total up the W/L record as well? Could be nice to show that in the search results instead of having to load each one individually. I think that's cheap since we're already iterating the matches, right? (I don't know C# and Mongo so this is outside my area of expertise).

marcoabreu:
> One thing to double check before deploy: EnsureIndexesAsync doesn't create an index on Teams.Players.BattleTag — the existing per-player history query already depends on it, so I assume prod has it manually. This endpoint rides the same index.

Just create it. EnsureIndexesAsync is new and was added by me as a mitigation for all those manual indizes that were added.

W3ChampionsStatisticService/Matches/MatchRepository.cs
```diff
@@ -140,6 +141,77 @@ public Task<long> CountFor(
...
+        string search,
+        int season,
+        GateWay gateWay = GateWay.Undefined,
+        GameMode gameMode = GameMode.Undefined,
+        int limit = 10)
+    {
+        var mongoCollection = CreateCollection<Matchup>();
+        var builder = Builders<Matchup>.Filter;
+        var playerMatchesFilter = builder.Where(m => m.Teams.Any(team => team.Players.Any(player => player.BattleTag == playerId)))
+            & builder.Where(m => m.Season == season)
```
marcoabreu, line 173: Is season not optional?

W3ChampionsStatisticService/Matches/MatchRepository.cs
```diff
@@ -140,6 +141,77 @@ public Task<long> CountFor(
...
+        string playerId,
+        string search,
+        int season,
+        GateWay gateWay = GateWay.Undefined,
+        GameMode gameMode = GameMode.Undefined,
+        int limit = 10)
+    {
+        var mongoCollection = CreateCollection<Matchup>();
+        var builder = Builders<Matchup>.Filter;
+        var playerMatchesFilter = builder.Where(m => m.Teams.Any(team => team.Players.Any(player => player.BattleTag == playerId)))
```
marcoabreu: playerId vs BattleTag?

What's playerId? Weird naming

W3ChampionsStatisticService/Matches/MatchRepository.cs
```diff
@@ -140,6 +141,77 @@ public Task<long> CountFor(
...
+            & builder.Where(m => m.Season == season)
+            & builder.Where(m => gateWay == GateWay.Undefined || m.GateWay == gateWay);
+
+        // An empty search matches everyone, i.e. returns the most played opponents.
+        var opponentFilter = new BsonDocument
```
marcoabreu: If no battletag has been specified, then the filter should be omitted? Otherwise this would be an expensive (or wrong??) regex? Like an empty regex matches nothing?

W3ChampionsStatisticService/Matches/OpponentInfo.cs
marcoabreu: Agree

DanyPell:
done in dfb8231 — added `{ Teams.Players.BattleTag: 1, Season: -1 }` compound to EnsureIndexesAsync, so the match-history list query gets it too instead of relying on whatever was created by hand.

W3ChampionsStatisticService/Matches/OpponentInfo.cs
DanyPell: added — the response now carries wins/losses per suggestion. You were right that it's basically free: the player's result gets resolved per match before the unwind, so it rides the existing group stage, still ~10 tiny docs out of mongo. Scoped to the game-mode filter exactly like matchCount, so the numbers always agree with the head-to-head you land on after clicking (checked on seeded data: dropdown says 6W - 6L, recap says Wins: 6 / Losses: 6). For allies it's the shared result, i.e. games won together.

W3ChampionsStatisticService/Matches/MatchRepository.cs
DanyPell: it is optional at the API level — the controller defaults to the latest season when it's not passed (same -1 convention as matches/search). The repo always scopes to one season on purpose: the UI is season-scoped anyway and it keeps the aggregation on the indexed battleTag+season subset.

W3ChampionsStatisticService/Matches/MatchRepository.cs
DanyPell: it's a battleTag, yeah — the query param is called playerId because that's what the whole matches API already uses for battleTags (matches/search?playerId=, ongoing/{playerId}), so renaming just this one would make the API inconsistent. Renamed it to battleTag internally (repo + service) so at least the code says what it is.

W3ChampionsStatisticService/Matches/MatchRepository.cs
DanyPell: other way around actually — an empty $regex matches everything, which was the intended "empty search = most played opponents" behavior (and it's cheap, the regex only ever runs on the already-unwound rows of the player's own season matches, never collection-wide). But agreed it read ambiguously, so the regex clause is now simply omitted when there's no search text — same behavior, self-explanatory.

PR #541, opened 2026-07-27
Add the admin job runner

francislavoie:
Lets operational one-offs be triggered from the admin UI instead of needing shell access to the environment. The immediate need is the MMR timeline backfill (separate PR), but the problem recurs — there is already a `W3ChampionsStatisticService.Tools/MapMetadataBackfill` command that has to be run out of band.

The first commit is a design doc (`docs/admin-job-runner.md`) written before the code; it carries the reasoning behind the decisions below and is the thing to read first.

**No jobs are registered in this PR** — `GET /api/admin/jobs` returns an empty list until the backfill lands.

### Scope

A job **runner**, not a scheduler. No cron, no retry/backoff, no chaining, no arbitrary parameters, no log streaming. Each is a reasonable follow-up; none is needed yet, and together they are the difference between a contained feature and a framework.

### Single instance

The service runs as one instance, and that assumption is what keeps this small: no queue, no polling loop, no heartbeat, no owner id. The API calls the runner directly.

Polling loops *do* exist in this codebase (`AsyncServiceBase.cs:48`), but those consume a continuous event stream — a different problem from "start this job now".

The assumption is load-bearing in exactly one place: the startup sweep that marks anything still `Running` as `Interrupted`. With a second replica, a booting replica would declare a job running elsewhere dead. That is commented where it lives, so scaling out is a conscious change rather than a silent breakage.

Starting a job is still a single atomic `FindOneAndUpdate` on its status — one line, and it is what stops two admins clicking Run at the same moment.

### Pause

There isn't one, deliberately. Cancel writes the checkpoint and Run resumes from it, so pause/resume is those two operations under different labels. A real in-memory pause would add a second, weaker resumption path — a job safe to interrupt is automatically safe to pause, but not the reverse — and could not survive a deploy anyway.

Resuming keeps the checkpoint, item count and accumulated runtime. `ActiveMs` banks each leg so a job left stopped overnight doesn't report twelve hours of runtime.

### Back-pressure

Jobs call `context.Pace()` between batches. Rather than a fixed duty cycle, this reads whether the system has headroom and runs flat out when it does, doubling the pause while pressure persists and halving it on recovery.

Two signals, sampled once a second:

- **Database.** This is a single mongod — no transactions or change streams anywhere in the solution, both of which need a replica set — so there is no replication lag to watch. The standalone equivalent is WiredTiger eviction pressure: replica lag is really a proxy for "writes are piling up faster than they can be durably absorbed", which here shows up as a growing dirty cache and, past a point, as mongod conscripting query threads into eviction. Those two counters need no threshold; any movement is the server complaining.
- **CPU.** Jobs share a process with the API, so one that pegs the CPU degrades every request the site serves even when the database is happy.

`FallbackDutyCycle` (25%) applies only when `serverStatus` cannot be read — most likely a user without `clusterMonitor`. Running flat out blind is the one genuinely dangerous option.

### Two traps worth knowing about

- The job runs in a DI scope the runner creates, not the request's. Inheriting the request scope disposes the job's services when the HTTP response completes.
- Progress writes are throttled in the context, so a job reporting per record does not spend more time writing progress than working.

### Test plan

- 24 tests. 13 against fakes (completion, failure, cancel, shutdown-vs-cancel, startup sweep, throttling, pacing) and 11 against real Mongo (claim/force/reset semantics, resume, duration across a pause).
- `PressureProbeTests` runs against a real server rather than a fixture, because WiredTiger statistic names are not a stable API and have moved between releases — if they move again the probe silently reports no pressure and jobs run flat out.
- Pre-existing `Mongo2Go` failures are unrelated.

### Open question for reviewers

Which permission each job should require. `RequiredPermission` is a property of the job, so individual jobs can be tightened later without touching the runner; `season-reset` is the likeliest candidate for one of its own. The three pressure thresholds are also judgement calls and want tuning against real load — the counter signals should carry most of the weight in practice.

### Not verified

The pressure thresholds have never been exercised under real load.

### DIFF
--- a/W3ChampionsStatisticService/Admin/Jobs/AdminJobsController.cs
+++ b/W3ChampionsStatisticService/Admin/Jobs/AdminJobsController.cs
@@ -1,0 +1,130 @@
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Net.Http.Headers;
+using W3C.Contracts.Admin.Permission;
+using W3C.Domain.Common.Services;
+using W3C.Domain.Tracing;
+using W3ChampionsStatisticService.WebApi.ActionFilters;
+
+namespace W3ChampionsStatisticService.Admin.Jobs;
+
+/// <summary>
+/// Manual trigger for operational jobs, so one-offs like the MMR timeline backfill do
+/// not need shell access to the environment. See docs/admin-job-runner.md.
+/// </summary>
+[ApiController]
+[Route("api/admin/jobs")]
+[Trace]
+public class AdminJobsController(
+    IEnumerable<IAdminJob> jobs,
+    IAdminJobRepository jobRepository,
+    AdminJobRunner runner,
+    IAuditLogService auditLogService) : ControllerBase
+{
+    private const string AuditCategory = "JOB";
+
+    private readonly List<IAdminJob> _jobs = jobs.OrderBy(j => j.Name).ToList();
+
+    [HttpGet]
+    [BearerHasPermissionFilter(Permission = EPermission.Jobs)]
+    public async Task<IActionResult> GetJobs()
+    {
+        var states = (await jobRepository.LoadAll()).ToDictionary(j => j.Id);
+
+        return Ok(_jobs
+            .Select(job => AdminJobDto.From(job, states.GetValueOrDefault(job.Key)))
+            .ToList());
+    }
+
+    [HttpPost("{key}/run")]
+    [BearerHasPermissionFilter(Permission = EPermission.Jobs)]
+    public async Task<IActionResult> RunJob(
+        string key,
+        string battleTag,
+        [FromQuery] bool force = false,
+        [FromQuery] bool reset = false)
+    {
+        var job = _jobs.FirstOrDefault(j => j.Key == key);
+        if (job == null)
+        {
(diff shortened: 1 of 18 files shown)


W3ChampionsStatisticService/Admin/Jobs/AdminJobsController.cs
```diff
@@ -0,0 +1,130 @@
...
+    private async Task<AdminJobDto> LoadDto(IAdminJob job) =>
+        AdminJobDto.From(job, await jobRepository.Load(job.Key));
+
+    /// <summary>
+    /// A job may require more than <see cref="EPermission.Jobs"/>, which is all the
+    /// filter attribute can express - it is static, and the job is only known per
+    /// request. Re-reads the caller's token rather than trusting anything from the
+    /// route.
+    /// </summary>
+    private bool CallerHasJobPermission(IAdminJob job)
```
marcoabreu, line 119: Isn't this redundant to [BearerHasPermissionFilter(Permission = EPermission.Jobs)]?

francislavoie: The design left it open to have a permission per job so we could have more finer-grained permissions instead of just one for everything (the job class would declare which permission it wants)

I can strip that out if you think that's over-engineered.

marcoabreu: Okay yeah makes sense.

PR #542, opened 2026-07-27
Add the MMR timeline backfill job

francislavoie:
> **Stack:** #540 → #542 → #543. Each PR's base is the one before it, so every diff shows only its own change. Merge in order.
>
> **Draft.** Its diff is against #540, so it shows only the job itself. The admin job runner it depends on merged in #541 and now comes from `master`; this branch no longer carries it.
>
> **#543 sits on top and is gated on this having run, not merely merged.**

Rebuilds `PlayerMmrRpTimeline` from the match events that produced it, so historical entries carry the Rd, Games and DailyMaxMmr the live handler now records. Until this runs, the lifetime endpoint withholds peaks for every player whose history predates those fields — which is nearly all of them.

### Source: `MatchFinishedEvent`, not `Matchup`

`Matchup` looked like the obvious source and is what the design doc originally assumed. It's subtly wrong: `PlayerOverviewMatches.OldRankDeviation` is the RD from *before* the match (`Matchup.cs:306` sets it from `w.mmr.rd`), while the live handler stores the RD from *after* (`player.updatedMmr.rd`). Backfilling from `Matchup` would mark players as calibrating slightly longer than live-written entries do — a permanent seam between backfilled and live history, in the exact field the feature depends on.

The events carry the handler's exact inputs, and the handler's exclusions (arranged teams, missing `updatedMmr`) are applied for the same reason. Coverage is equivalent: `Matchup` is *derived from* these events, so it cannot have deeper history. The events collection appears on the second day of the repository (`569f39c`, "saving of raw data working"), has no TTL index, is never deleted from, and has a sync path that fills gaps in.

Events are bulkier, so the query projects down to the five fields used.

### Batching

Day-batched and ranged on `_id`, whose leading bytes are a creation timestamp, so the always-present `_id` index serves it. Every match is read once in order and never looked up by player — which is why this needs **no index on `Teams.Players.BattleTag`**. Adding one would otherwise stall startup, since `MongoIndexInitializationService` awaits `EnsureIndexesAsync` and `createIndexes` doesn't return until the build completes.

A season would be too wide (~1M matches, ~15k players); a day is ~6k matches and flat regardless of how much history is processed.

### Two operational details

**It stops at yesterday (UTC), relative to whenever it runs.** Today is still being written by the live handler and rewriting a day underneath it would drop games landing mid-run. The corollary: **run it on a later UTC day than #540 is deployed on.** On the deploy day itself, entries the old handler wrote that morning are never rebuilt yet the document is still promoted — claiming games counts that were never computed.

**There is no schema-version bookkeeping.** An earlier revision marked each timeline as it was rebuilt and promoted the marks in one pass at the end, so the lifetime endpoint could tell verified history from unverified. That precision is not worth its cost for a stats page, so the field, the marker and the promotion pass have all been removed. The job now rebuilds days and stops.

### This rewrites existing history

Rebuilding recomputes each day under today's rules, so days where a player didn't finish on their high point will move. **This is intended, not tolerated:** the alternative leaves each timeline half-computed one way and half the other.

### Fixes from review

- **Writes are conditional on the revision each document was read at.** The live handler rewrites these same documents whole while the job runs; without a guard whichever wrote second silently destroyed the other's work. On a clash the whole day is redone, which is free because rebuilding a day is already idempotent.
- **The pipeline's own filters are applied**, not just the handler's. The job reads the event collection raw, so it has to repeat what `MatchFinishedReadModelHandler` does before the handler ever sees an event: skip cancelled matches, and skip null/empty-battleTag players. The null case failed the job outright, and a resume would have retried the same day forever.

Only `CANCELED` is rejected, deliberately — `match.state` was an untyped int with no filter at all until #405 (2025-05-26), so events stored before then were folded into timelines whatever their state, and an absent field reads as `INIT`. Since a day is cleared before being rewritten, rejecting those would delete history rather than reproduce it.

### Test plan

- 13 tests covering day folding with peak and games count, the RD threshold, arranged-team exclusion, per-race separation, rerun convergence, rebuilding over old-handler data, leaving other days alone, leaving today alone, not claiming untouched timelines, checkpoint resume, and marker cleanup.
- These derive from `IntegrationTestBase`, whose `[SetUp]` drops a shared remote database, so **they have not been run locally since the rebase** — CI covers them. The rebase is build-verified only (`dotnet build`, 0 errors).
- Pre-existing `Mongo2Go` failures are unrelated.

### Not verified — please read before running

**This has never been run at real data volume.** The scale reasoning is the design doc's, not measured. It rewrites production history, so I'd want a dry run against a restored snapshot first — both for timing and to see what `LastPressureReason` reports, which is also the data needed to tune the runner's pressure thresholds.

Worth confirming event coverage against prod once, comparing the oldest `_id` in `MatchFinishedEvent` and `Matchup`.

### DIFF
--- a/W3ChampionsStatisticService/Admin/Jobs/AdminJobServiceExtensions.cs
+++ b/W3ChampionsStatisticService/Admin/Jobs/AdminJobServiceExtensions.cs
@@ -19,6 +19,8 @@ public static class AdminJobServiceExtensions
         services.AddHostedService(sp => sp.GetRequiredService<AdminJobRunner>());
 
         // Job registrations. Discovered by the runner and the controller as IAdminJob.
+        services.AddInterceptedScoped<IAdminJob, PlayerMmrRpTimelineBackfillJob>();
+
         return services;
     }
 }
--- a/W3ChampionsStatisticService/Admin/Jobs/PlayerMmrRpTimelineBackfillJob.cs
+++ b/W3ChampionsStatisticService/Admin/Jobs/PlayerMmrRpTimelineBackfillJob.cs
@@ -33,7 +33,9 @@ public class PlayerMmrRpTimelineBackfillJob(MongoClient mongoClient) : IAdminJob
 
     public string Description =>
         "Recomputes the MMR/RP timeline from match history so lifetime peaks can be shown. " +
-        "Rewrites historical entries, which shifts some existing charts. Resumable; runs for hours.";
+        "Rewrites historical entries, which shifts some existing charts. Resumable; runs for hours. " +
+        "Covers everything up to yesterday (UTC) - do not run this on the same UTC day the " +
+        "timeline handler was deployed, or that day's part-old entries will be marked as rebuilt.";
 
     public bool RequiresConfirmation => true;
 
@@ -59,8 +61,14 @@ public class PlayerMmrRpTimelineBackfillJob(MongoClient mongoClient) : IAdminJob
             return;
         }
 
-        // Yesterday, not today: today is still being written by the live handler, and
-        // rewriting a day underneath it would drop games that land mid-run.
+        // Yesterday relative to whenever this runs, not to any deploy: today is still
+        // being written by the live handler, and rewriting a day underneath it would
+        // drop games that land mid-run.
+        //
+        // The corollary is operational: run this on a later UTC day than the one the
+        // timeline handler was deployed on. Run it on the same day and the entries the
+        // OLD handler wrote earlier that day are never rebuilt, yet the document is
+        // still promoted to the current schema version, claiming they were.
         var lastDay = DateOnly.FromDateTime(DateTime.UtcNow.Date).AddDays(-1);
         var day = ReadCheckpoint(context) ?? firstDay.Value;
 
--- a/W3ChampionsStatisticService/Admin/Jobs/PlayerMmrRpTimelineBackfillJob.cs
+++ b/W3ChampionsStatisticService/Admin/Jobs/PlayerMmrRpTimelineBackfillJob.cs
@@ -96,11 +96,46 @@ public class PlayerMmrRpTimelineBackfillJob(MongoClient mongoClient) : IAdminJob
         await PromoteSchemaVersion(context, cancellationToken);
     }
 
+    /// <summary>How many times to redo a day whose documents moved underneath it.</summary>
+    private const int MaxRebuildAttempts = 5;
+
     /// <summary>
     /// Rebuilds every timeline that has a match on this day, and returns how many were
     /// written.
+    /// <para>
+    /// The live handler writes these same documents while this job runs, and both
+    /// replace them whole - so every write is conditional on the revision it was read
+    /// at. On a clash the whole day is redone rather than the individual document
+    /// retried: rebuilding a day is idempotent (the day is cleared and re-derived from
+    /// the events), so redoing it is both simpler and exactly equivalent.
+    /// </para>
     /// </summary>
     private async Task<int> RebuildDay(DateOnly day, CancellationToken cancellationToken)
+    {
+        for (var attempt = 1; ; attempt++)
+        {
+            var written = await TryRebuildDay(day, cancellationToken);
+            if (written.HasValue)
+            {
+                return written.Value;
+            }
+
+            if (attempt >= MaxRebuildAttempts)
+            {
+                throw new InvalidOperationException(
+                    $"Could not rebuild {day:yyyy-MM-dd} after {MaxRebuildAttempts} attempts; its timelines kept being written concurrently.");
+            }
+
+            Log.Information("Timelines for {Day} changed while rebuilding, redoing the day ({Attempt}/{MaxAttempts})",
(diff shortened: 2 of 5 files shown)


francislavoie:
## Spec: the lost-update race between the backfill and the live handler

Found independently by two reviewers. **Not fixed in this PR** — it needs a decision, and the fix is small once the direction is chosen.

### The mechanism

Both writers do load → mutate → **replace the whole document**, with no concurrency control:

- Live handler: `PlayerRepository.Upsert` → `FindOneAndReplaceAsync` (`MongoDbRepositoryBase.cs:63-71`)
- Backfill: `RebuildDay` → `ReplaceOneModel` (`PlayerMmrRpTimelineBackfillJob.cs:124-130`)

Neither checks that the document is unchanged since it was read. So for any player who is *actively playing while the job runs*:

**Backfill write lands second** — today's live entry is silently dropped. The backfill never repairs today by design, and the handler's watermark has already advanced, so it is never rewritten. A permanently missing day.

**Live write lands second** — the rebuilt day is reverted to the handler's stale snapshot, **including the `BackfillPending` marker being stripped**. If that player has a later historical day, the backfill re-marks the document, and `PromoteSchemaVersion` then stamps `SchemaVersion = 1` over a day still in old last-game-only shape.

That second case is the one that matters. #543 gates only on `SchemaVersion >= 1`, and an absent `Rd` reads as "settled" — so **a calibration-era rating gets published as a lifetime peak.** That is precisely the outcome the withholding design exists to prevent.

### Why the exposure is real

Per event the window is milliseconds. But the job runs for hours, touches roughly one timeline write per player-day across all history, and concentrates on *active* players — who are both the most likely to be playing concurrently and the most likely to have their profile viewed.

### Options

**A. Targeted per-day update instead of whole-document replace.** `$pull` the day being rebuilt, `$push` the new entries, `$set` the marker — one atomic update touching only that day. The live handler writes a different day (today), so the two stop colliding entirely.
*Best correctness. Largest change: `RebuildDay` currently rewrites the entry list wholesale, and the `$push` needs `$sort`/`$position` to preserve ordering.*

**B. Optimistic concurrency on the replace.** Add a version or `_updated_at` field, make both writers include it in the replace filter, and retry on zero matched. Protects against any future concurrent writer, not just these two.
*Moderate change, touches the shared `MongoDbRepositoryBase` path — so it needs care not to disturb other repositories.*

**C. Do not run the backfill while the handler is live.** `START_HANDLERS=false` on the instance running the job, or run it during a maintenance window.
*No code. Operationally awkward: the job takes hours, and the handlers are the live read-model pipeline.*

**D. Accept, and repair afterwards.** Add a verification pass that finds `SchemaVersion = 1` timelines containing entries that cannot be version-1 shaped, and demote them.
*Detects rather than prevents; still leaves the missing-day case unfixed.*

### Recommendation

**B**, with **C** as the immediate mitigation for the first run. B is the smaller change of the two real fixes and protects against writers we have not thought of; A is strictly better but rewrites the hot path of the job. Doing C for the first run means the first pass — the one that touches every timeline in history — is not exposed at all, and B can land before any subsequent re-run.

Whichever is chosen, it belongs in this PR or before it, not in #543: by the time #543 can observe the damage, the timeline has already been promoted.

francislavoie:
## Option B implemented

Optimistic concurrency, split across the two branches that own each writer:

- **#540** — `Revision` on the document, `TryUpsertPlayerMmrRpTimeline`, and the handler's retry loop.
- **#542** (this PR) — the backfill's bulk write is now filtered on the revision each document was read at.

### Notes on the implementation

**`expectedRevision` is nullable, not defaulted to 0.** Every document predating the field deserializes as revision 0, so "no document was loaded" and "a document at revision 0" are different states. Collapsing them would silently overwrite a concurrent creator instead of detecting it. That path inserts rather than upserts, and treats the duplicate key as the conflict it is.

**The handler retries the whole read-modify-write, not just the write.** The merge depends on what the day already holds, so reapplying it to a stale copy would reinstate the state the other writer replaced. After five attempts it throws and hands the event back to the pipeline — which is safe now that `LastProcessedMatchId` makes re-folding a no-op for players already written.

**The backfill redoes the whole day on a clash** rather than retrying individual documents. `RebuildDay` already clears the day and re-derives it from the events, so redoing it is idempotent and exactly equivalent — and much simpler than working out which of a few thousand bulk writes missed.

**Not put in `MongoDbRepositoryBase`.** That path is shared by every repository in the service and only this collection has two writers.

### Verification

Run against a local MongoDB (`mongo:8.0.19` — note `mongo:8.0` will not start on kernel 6.19+, SERVER-121912), then the connection string reverted. **48 tests pass** across `PlayerTests` and the backfill tests, including three new ones:

- a write against a stale revision is refused and the concurrent write survives
- a second creation is reported as a conflict rather than overwriting the first
- a document whose revision moves behind the job's back is rebuilt on top of the concurrent write rather than reverting it

### Still worth doing before the first run

The mitigation from option C is no longer required, but the remaining open items from the review stand: the `±2h` `_id` padding gap, the unpaced `PromoteSchemaVersion` mega-write, and the one-off production query on `MatchFinishedEvent` counts by state (see the filter commit).

PR #544, opened 2026-07-27
feat(search): one search core with ladder-context relevance and rank-in-context lookup

MarkWind85:
One search core serves every place the site looks a player up by name. `GET api/players/global-search` gains optional ladder-context relevance, and a new `POST api/ladder/ranks-for-players` annotates exact battleTags with their rank in one season, gateway and game mode. **Deploy this before the companion website PR** — the reverse order breaks: the old backend has no `POST` route there and answers 405 with an empty body, which wedges the new frontend's search spinner.

The full architecture reference ships in this PR — `docs/PLAYER_SEARCH_CONSOLIDATION.md` — with the deeper rationale for everything summarized below, including the retirement plan for the legacy endpoints.

```mermaid
flowchart LR
  H["Header"] --> CORE
  P["Player picker"] --> CORE
  R["Rankings"] -->|"+ season, gateway, gameMode"| CORE
  CORE["api/players/global-search<br/>relevance · cursor-paged"] --> REFS[/"lightweight player refs<br/>battleTag · name · seasons · avatar"/]
  REFS -->|"ladder only"| ENR["api/ladder/ranks-for-players"]
  ENR --> RANK[/"mmr · W-L · league · rankNumber<br/>absent = unranked"/]
```

## What changes when: live on deploy vs. dormant until the flag

Two different things ship here, and they activate at different moments:

**Live the moment this deploys** — independent of the website:

- **Country, clan and chat rank lists become complete.** These lookups match only the two stored member fields today, so the third and fourth members of an arranged team are missing from country rankings and clan rank lists. They now match the full member list. Response shapes are unchanged; the row *sets* grow — that growth is the fix.

- **`global-search` rejects unsearchable terms.** The route currently accepts any term: an empty search matches the entire player directory and pages it out anonymously, and a null search is a 500. It now requires three letters-or-digits (400 otherwise). This is a deliberate behaviour change on a shipped endpoint, and its one external caller is verified safe: launcher-e gates its own input at three characters in two places and treats any non-200 as an empty result. Counting letters and digits rather than raw length closes a real bypass — matching is collation-aware, so a term of three zero-width characters passes a length check and matches every battleTag.
-  **The `Rank` storage change and its backfill** (next section).

**Dormant until the website flag flips** — the new capability, called by nobody at deploy time:

- Ladder-context relevance in `global-search`: given `season`/`gateWay`/`gameMode`, players ranked on that ladder come first, ordered by ranking points (league ids follow creation order, so divisions added mid-season sort wrong by id). Sending a partial context is a 400 — the relevance id doubles as the pagination cursor, and a caller that dropped a parameter between pages would silently get nothing. The parameters are new; no existing caller can hit either behaviour.
- `POST api/ladder/ranks-for-players`: rank lookup for up to 50 exact battleTags — a request-size bound on an anonymous route, not a result cap.
- **Context-free responses are unchanged**: same ordering, same relevance-id format, pinned by tests. launcher-e and the header keep receiving exactly what they receive today.

## Why the context must reach the core

Filtering a name-ordered page afterwards cannot work: how well a name matches says nothing about being ranked. Measured on the 2022 dump: `man` on season 13 / EU / 1v1 matches 766 players, 55 of them ranked on that ladder — and the first name-ordered page of 20 contains **none of them**. Ordered by standing, page one holds 20 ranked players and the rest follow as the list scrolls.

## Schema change: the `Rank.MemberIds` list

A rank row now names **every** member of the entry; the long-standing `Player1Id`/`Player2Id` keep holding the first two. The list exists for the lookups alone and stays out of API responses (`[JsonIgnore]`, pinned by a contract test), so no rank-returning endpoint changes shape.

**What the backfill repairs** — rows written before the field existed. Real examples from the
2022-11-27 prod dump (173,568 rows across 14 seasons, filled in 13.3s, every row verified against its expected outcome):

| Row (`_id`) | Before: `Player1Id`/`Player2Id` | After: `MemberIds` |
|---|---|---|
| `13_Teror#21981@20_GM_1v1_OC` | `Teror#21981` / null | `[Teror#21981]` |
| `13_KraV#21403@20_StarBuck#2732@20_GM_2v2_AT` | `StarBuck#2732` / `KraV#21403` | `[StarBuck#2732, KraV#21403]` |
| `13_FEWA418#1895@20_HumanKingTom#11421@20_Spark#12912@20_GM_4v4_AT` | `HumanKingTom#11421` / `Spark#12912` | `[HumanKingTom#11421, Spark#12912, FEWA418#1895]` |
| `13_InSaNe#12987@20_JmA#11680@20_Nymph#11746@20_RandomsTyLe#1383@20_GM_4v4_AT` | `InSaNe#12987` / `RandomsTyLe#1383` | `[InSaNe#12987, JmA#11680, Nymph#11746, RandomsTyLe#1383]` |
| `13_Seal#21311@20_Wing#21498@20_GM_2v2_AT` — no `PlayerOverview` exists | `Wing#21498` / `Seal#21311` | `[Wing#21498, Seal#21311]` (from the two stored fields) |

The last two members of the fourth row are the point: a row can store only two members, so the rest of the team exists only inside the `_id` string and the `PlayerOverview` — searching them by name finds no rank until the backfill has run. The dump holds 829 rows like that (536 with three members, 293 with four), all in seasons 9–13.

**How the repair ships.** It is the `rank-member-ids-backfill` admin job — the first job on the
#541 runner, surfaced by the admin Jobs page (w3champions/website#1110) — run once per
environment. Until it has run, the new member lookups miss the old rows; the country, clan and chat pages keep working through their old two-field match, which is removed once the legacy search retires. The operator playbook — how to trigger, how to verify, what can fail — is in `docs/PLAYER_SEARCH_CONSOLIDATION.md` in this diff.

## Scope: directory search only

Directory search asks "which players match this name" — the header, the player picker and the
rankings page all ask it, and this core serves all three. "Who have I played" is match-scoped,
carries per-pair data, keeps its own endpoint, and is served by w3champions/website#1103. The two consolidations complement each other.

## External consumers, pinned by contract tests

| Client | Endpoint | Contract held |
|---|---|---|
| launcher-e (Tauri, active) | `global-search` | sends `pageSize=10` and infers a further page from receiving exactly 10 — honoured, and every field it reads is asserted |
| in-game UI bundle (deprecated Electron launcher, still live) | `ladder/search` | exact call shape and its response — one array holding the ranked block plus zeroed rows for unranked name matches — plus `MemberIds` absent from the serialized model |

### DIFF
--- a/W3ChampionsStatisticService/Admin/Jobs/AdminJobServiceExtensions.cs
+++ b/W3ChampionsStatisticService/Admin/Jobs/AdminJobServiceExtensions.cs
@@ -1,5 +1,6 @@
 using Microsoft.Extensions.DependencyInjection;
 using W3ChampionsStatisticService.Extensions;
+using W3ChampionsStatisticService.Ladder;
 
 namespace W3ChampionsStatisticService.Admin.Jobs;
 
@@ -19,6 +20,7 @@ public static class AdminJobServiceExtensions
         services.AddHostedService(sp => sp.GetRequiredService<AdminJobRunner>());
 
         // Job registrations. Discovered by the runner and the controller as IAdminJob.
+        services.AddInterceptedScoped<IAdminJob, RankMemberIdsBackfillJob>();
         return services;
     }
 }
--- a/W3ChampionsStatisticService/Ladder/LadderController.cs
+++ b/W3ChampionsStatisticService/Ladder/LadderController.cs
@@ -44,6 +44,21 @@ public class LadderController(
         return Ok(playerRanks);
     }
 
+    [HttpPost("ranks-for-players")]
+    public async Task<IActionResult> GetRanksForPlayers([FromBody] RanksForPlayersRequest request)
+    {
+        if (request?.BattleTags == null || request.BattleTags.Count == 0)
+        {
+            return Ok(new List<RankInContext>());
+        }
+        if (request.BattleTags.Count > RanksForPlayersRequest.MaxBattleTags)
+        {
+            return BadRequest($"battleTags must contain at most {RanksForPlayersRequest.MaxBattleTags} entries.");
+        }
+        var ranks = await _rankRepository.LoadRanksForPlayers(request.BattleTags, request.Season, request.GateWay, request.GameMode);
+        return Ok(ranks.Select(r => r.ToRankInContext()).ToList());
+    }
+
     [HttpGet("{leagueId}")]
     public async Task<IActionResult> GetLadder([FromRoute] int leagueId, int season, GateWay gateWay = GateWay.Europe, GameMode gameMode = GameMode.GM_1v1)
     {
--- a/W3ChampionsStatisticService/Ladder/PlayerLadderStanding.cs
+++ b/W3ChampionsStatisticService/Ladder/PlayerLadderStanding.cs
@@ -6,8 +6,8 @@ namespace W3ChampionsStatisticService.Ladder;
 /// Where a player sits on one ladder — the whole of what ordering a search by ladder standing needs.
 /// </summary>
 /// <remarks>
-/// Deliberately narrower than <see cref="Rank"/>: the search core reads League and RankNumber and
-/// nothing else, so hydrating the joined PlayerOverview on every hit would be payload it never opens.
+/// Deliberately narrower than <see cref="Rank"/>: the search core reads RankingPoints and nothing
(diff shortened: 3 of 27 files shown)


MarkWind85:
Waiting on https://github.com/w3champions/website-backend/pull/541 - will need to do some rework and additional testing when jobs backfill is live on test

PR #545, opened 2026-07-28
fix: restore friends/presence for released launchers (temporary JWT fallback on wb hub)

DanyPell:
## What broke

The #538 chat-revamp cutover made `WebsiteBackendHub` ticket-only ("NO raw-JWT fallback"), designed to ship lock-step with a forced launcher update. That launcher release never shipped — the latest is **v1.6.5 (June 16)**, which predates the ticket flow entirely. The cutover image (build 13352, July 7) sat unused until the recent statistic-service deploy put it live; from that moment every released launcher presents a raw JWT, gets `AuthorizationFailed`, and loses friends/presence. (This is why it "worked yesterday": merge was July 7, the deploy was recent.)

## Fix

Ticket-first with a **temporary** raw-JWT fallback: try `TryConsume` (updated launchers), fall back to `GetUserByToken(accessToken, false)` — the exact pre-cutover validation — for released launchers. `GetUserByToken` throws on invalid tokens (every REST filter wraps it in try/catch), so the fallback catches too: an updated launcher whose single-use ticket was already consumed rejects cleanly instead of surfacing a hub exception.

Deploying this restores friends for the entire playerbase immediately, **no launcher update or rollback required**. Once a launcher release containing the ticket mint (launcher-e #833) has shipped and its forced-update rollout completed, remove the fallback in a follow-up — that completes the original cutover as designed.

## Tests

3 new: legacy raw JWT connects via fallback; a valid ticket never consults the fallback (pins ordering); a token invalid on both paths rejects cleanly. The auth-service test mock **throws** on invalid tokens, mirroring the real service contract, so every rejection test exercises the catch path. 38/38 hub + 17/17 session tests green.

## Notes for reviewers

- Known cost while the bridge lives: released launchers keep the pre-cutover posture (JWT in the WS URL query string) — the thing tickets fix. Accepted temporarily; the removal follow-up closes it.
- `TicketStore` is single-instance in-memory by design — fine on one replica; if statistic-service is ever scaled out, tickets will fail randomly (mint on one instance, WS lands on another). Worth confirming prod topology.
- chat-service has the identical undeployed cutover (#33) and only works today because its server side never deployed. A matching bridge PR for chat-service follows so its next deploy is safe in any order.

### DIFF
--- a/W3ChampionsStatisticService/Hubs/WebsiteBackendHub.cs
+++ b/W3ChampionsStatisticService/Hubs/WebsiteBackendHub.cs
@@ -29,7 +29,8 @@ public class WebsiteBackendHub(
     TracingService tracingService,
     IBattleTagResolver battleTagResolver,
     IRelationshipChangeNotifier relationshipChangeNotifier,
-    ITicketStore ticketStore
+    ITicketStore ticketStore,
+    IW3CAuthenticationService authenticationService
 ) : Hub
 {
     static WebsiteBackendHub()
@@ -59,6 +60,7 @@ public class WebsiteBackendHub(
     private readonly IBattleTagResolver _battleTagResolver = battleTagResolver;
     private readonly IRelationshipChangeNotifier _relationshipChangeNotifier = relationshipChangeNotifier;
     private readonly ITicketStore _ticketStore = ticketStore;
+    private readonly IW3CAuthenticationService _authenticationService = authenticationService;
 
 
     [NoTrace]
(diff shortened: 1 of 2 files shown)


DanyPell:
Good catch — the `search-opponents` endpoint (and a related test-mock commit) were unrelated local commits that were accidentally included when this branch was cut from a local checkout instead of `origin/master`. Rebased them out; the PR now contains exactly one commit touching only `WebsiteBackendHub.cs` and its tests.

PR #547, opened 2026-08-05
fix: stop read model handlers wedging forever on poison events

marcoabreu:
## Problem

Two read model handlers have been wedged in production, each throwing every 5 seconds — ~50,000 errors/day combined. `LagReportMatchCanceledHandler` has been stuck since **~2021**; `LagReportMatchFinishedHandler` for months.

```
Some Readmodelhandler is dying: ... deserializing the match property of MatchCanceledEvent
  ... Cannot deserialize a 'Race' from BsonType 'Null'.
Error processing "MatchFinishedEvent" ... Received match with illegal state CANCELED
  within the MatchFinishedReadModelHandler
```

Neither is a transient failure — both are permanent conditions that retrying can never resolve, so the handlers never advance past them.

## Changes

**1. Null-tolerant `race` — the actual blocker** (`MatchEventDtos.cs`)

`UnfinishedMatchPlayer.race` was a non-nullable `Race` fed BSON `null` by old documents. This throws inside `Load<T>()`, i.e. before the per-event try/catch can even see it, so it has to be tolerated at the model level.

Uses the private-backing-property idiom already established by `Match.endTime` in this same file; both a missing and a null race now map to `Race.RnD` (0), which is exactly what a missing field already did. The public property stays non-nullable `Race`, so the ~20 consuming handlers are untouched. Verified nothing queries or sorts on `.race` in Mongo, and the write path stays symmetric.

**2. Events that aren't the handler's business are filtered, not fatal**

A `MatchFinishedEvent` carrying a CANCELED match is not a failure — it is simply not the finished handler's concern, and processing it would actively be wrong (it would add a non-counting game to the stats read model). It also can never become FINISHED later, so throwing just wedges the handler on a permanent condition.

`void ValidateMatchState` (throws) becomes `bool ShouldProcessEvent`. Returning false skips the inner handler and advances the watermark. CANCELED-in-finished logs at Information (known and potentially frequent — must not drown out real warnings); genuinely unexpected states log at Warning. No failure counting is involved anywhere in this path.

**3. Pre-existing watermark bug fixed in passing**

`ProcessMatchEvent` returned the stale `lastVersion` after persisting a new one, so a later season bump in the same batch re-persisted a one-event-behind watermark and re-delivered already-handled events after a crash. All advance sites now go through `AdvancePast()`. Happy to split this out if you'd rather review it separately.

## Retry behaviour is unchanged

Genuine failures still `throw` and are retried forever — deliberately. An earlier revision of this PR added an auto-skip after N failures; that was removed at review, and the test `FailingEvent_IsRetriedAndDoesNotAdvanceTheWatermark` now guards the forever-retry contract explicitly so it can't be eroded later.

## Deploy notes

- Fix 1 clears the `race` blocker, so `LagReportMatchCanceledHandler` unwedges on deploy and **replays ~5 years of `MatchCanceledEvent`s**, calling flo-stats for each. This is expected per our decision to let it replay; worth watching flo-stats load, and the watermark can be fast-forwarded by hand if it bites.
- Fix 2 unwedges `LagReportMatchFinishedHandler`, which is currently stuck on event `6799240b2c3163adbb721c48` (a real 1v1 match from Jan 2025 that finished and was then canceled).

## Known limitation

The `race` fix is at the model level precisely because a whole-batch failure inside `Load<T>()` bypasses every per-event guard. Any *future* undeserializable document would wedge a handler the same way, and would need the same treatment — making `Load()` itself isolate bad documents one-by-one is deliberately out of scope here.

## Testing

TDD throughout; the `race` test failed RED with the literal production error. New `MatchEventDtoBsonTests` covers null, missing and normal races. Existing tests asserting `ThrowsAsync<InvalidOperationException>` were updated to assert skip-and-advance.

```
dotnet build                       0 errors (1 pre-existing unrelated CS1998 warning)
dotnet format --verify-no-changes  clean
ReadModel fixtures                 16/16 passed
LagReports fixtures                60/60 passed
Matchups+PlayerProfiles+Statistics 126/126 passed (race regression check)
```


W3ChampionsStatisticService/Ports/IVersionRepository.cs
```diff
@@ -5,7 +5,17 @@ namespace W3ChampionsStatisticService.Ports;
...
+    /// <summary>
+    /// Records that <paramref name="eventId"/> failed and returns how often it has failed in a row.
+    /// The count has to be persisted because the read model handlers are transient: a fresh instance
+    /// is resolved for every retry cycle, so an in-memory counter would reset before it ever trips.
+    /// A different event id than the last recorded one restarts the count at 1.
+    /// </summary>
+    Task<int> RecordEventFailure<T>(string eventId);
 }
 
 public class HandlerVersion(string version, int season, bool isStopped)
```
marcoabreu: We do not want to just skip events that have a failure. It's intentional that it loops forever if we have a failure. The idea is that a human manually resolves - never an auto-skip after X failures!!

W3ChampionsStatisticService/ReadModelBase/MatchEventReadModelHandler.cs
```diff
@@ -19,6 +19,17 @@ public abstract class MatchEventReadModelHandler<TEvent, THandler>(
     where TEvent : MatchmakingEvent
     where THandler : class
 {
+    /// <summary>
+    /// How often the same event may fail before it is written off as poisonous and skipped.
```
marcoabreu: Again, we do not want this!

marcoabreu:
Understood — the whole bounded-retry/auto-skip mechanism is gone as of 80f8ecd.

`RecordEventFailure`, `MaxEventFailuresBeforeSkip`, the `FailingEventId`/`FailureCount` fields and their tests are all removed. `IVersionRepository.cs` and `VersionRepository.cs` are now byte-identical to master, and the catch block is back to the original unconditional `throw;`. Failures loop forever again, for a human to resolve.

I kept the retry test and renamed it `FailingEvent_IsRetriedAndDoesNotAdvanceTheWatermark` so the forever-retry contract is now explicitly pinned by a test.

Two things left in, both of which you confirmed or which are separable:

1. **`ShouldProcessEvent` / CANCELED** — this is not a failure path and involves no failure counting. A canceled match simply isn't the finished handler's business, and processing it would be wrong (it'd add a non-counting game to the stats read model). Per your call, it logs and advances rather than throwing.
2. **The `AdvancePast` watermark fix** — a pre-existing bug where `ProcessMatchEvent` returned a stale `lastVersion`, so a season bump later in the same batch re-persisted a one-event-behind watermark and re-delivered events after a crash. Unrelated to the skip mechanism; say the word if you'd rather I split it into its own PR.

PR #548, opened 2026-08-06
Cancelled matches moderation page: backend

francislavoie:
Backend half of the **Cancelled Matches** moderation page — matches that ended with no result, so no winner, no score, no MMR change.

The upstream endpoint and FLO slot index it consumes landed in w3champions/matchmaking-service#866, which is now **merged** — so this is no longer blocked. (It returns 502 if the matchmaking service is older than that.)

## Why a proxy and not a read model

The obvious design — a `CanceledMatch` read model fed by `MatchCanceledEvent` — is wrong here, because that stream is lossy. `pushMatchCanceled` has exactly one call site (the timeout sweeper); the game-creation-failure path calls `cancelMatch` and pushes nothing, and doesn't emit a `MatchStartedEvent` either. Those matches would simply be missing.

Matchmaking's `Match` collection records every cancellation, so this proxies it. That also drops the backfill, the `HandlerVersions` cursor, the `START_HANDLERS` dependency, and a heal-removal handler — a healed match changes state and leaves the result set on its own.

## What's here

- **`GET api/admin/matches/canceled`** (`AdminMatchesController`), gated on `EPermission.Moderation`, proxying matchmaking via `MatchmakingServiceClient`.
- **Reuses the existing `Match` / `PlayerMMrChange` / `UnfinishedMatchPlayer` DTOs** rather than adding parallel ones. They already model this document; they only needed JSON mapping, since they were BSON-annotated and `MatchmakingServiceClient` deserialises with Newtonsoft — **without `[JsonProperty("_id")]` the match id silently came back null.** Verified safe: those DTOs are not returned by any controller (`MatchesController`'s `Ok(match)` returns `MatchupDetail`, which holds a `Matchup`).
- **Replay and chat logs now 404 properly when unavailable.** Previously `GetChatLogs` swallowed the status and returned 200-with-null, while `GenerateReplay` threw and became a 500. Both now return `null` for 404/410 only — any other upstream status logs and fails loudly, so a replay-service **outage stays distinguishable from a replay that was never recorded or has aged into DeepArchive**.
- **Moderators get 50 replay downloads/hour instead of 10**, partitioned per battletag rather than per IP. Cancelled matches always resolve to the strict bucket, because `CheckMatchAge` returns null for them. The daily number is unchanged, and `Math.Max` means the override can only ever raise a limit.
- **`GM_FOOTMEN_FRENZY` is now treated as an FFA mode.** ⚠️ This is a live behaviour change beyond the feature: its matchmaking definition sets `isAnonymous = true`, so flo masks names in game, but the backend was not treating it as FFA — ongoing Footmen Frenzy matches were exposing real battletags through `/api/matches/ongoing` and were lookupable by player. Anon-FFA Footmen Frenzy is returning, so this needs to be right. It reaches three pre-existing call sites (`MatchesController.cs:201`, `PlayersObfuscator.cs:12`, `Matchup.cs:160`), all anonymisation behaviours.

## Two bugs worth calling out, both caught in review

**The search parameter cannot be named `battleTag`.** `BearerHasPermissionFilter.cs:33` does `context.ActionArguments["battleTag"] = res.BattleTag;` unconditionally, so on any action it guards, a parameter with that name is silently replaced by the *acting moderator's* tag — the endpoint would have quietly returned only the moderator's own matches. It is `playerBattleTag`, with a regression test that fails if the name comes back. Unit tests could not have caught this: they invoke the controller directly and bypass the filter pipeline.

**The moderator rate limit was initially unreachable** from the page it exists for — the download button sent no `Authorization` header, so every request landed in the anonymous IP partition. Fixed in the companion frontend PR; noted here because the two halves have to agree.

## Testing

`dotnet build` clean, 45/45 passing across every test class this touches.

There is no dotnet SDK on this dev host, so `dotnet-docker.sh` is included — it runs the same SDK image the Dockerfile uses. It carries a warning worth repeating: **never run an unfiltered `dotnet test`**, because `IntegrationTestBase` drops and recreates the real shared `W3Champions-Statistic-Service` database on a remote host in `[SetUp]`.

Consequently the Mongo-backed handler tests could not run here. BSON back-compatibility of the DTO changes was verified with a targeted `BsonSerializer.Deserialize<Match>` test (fields present as dates, and absent) rather than the integration suite — worth confirming in CI.

The design spec and implementation plan are included under `docs/superpowers/`.

### DIFF
--- a/W3C.Domain/MatchmakingService/MatchmakingServiceClient.cs
+++ b/W3C.Domain/MatchmakingService/MatchmakingServiceClient.cs
@@ -114,6 +114,32 @@ public class MatchmakingServiceClient
         return null;
     }
 
+    public async Task<CanceledMatchesResponse> GetCanceledMatches(CanceledMatchesGetRequest req)
+    {
+        var url = $"{MatchmakingApiUrl}/admin/canceled-matches?page={req.Page}&itemsPerPage={req.ItemsPerPage}";
+
+        if (req.GameMode != GameMode.Undefined)
+        {
+            url += $"&gameMode={(int)req.GameMode}";
+        }
+
+        if (!string.IsNullOrEmpty(req.BattleTag))
+        {
+            url += $"&battleTag={HttpUtility.UrlEncode(req.BattleTag)}";
+        }
+
+        var request = new HttpRequestMessage(HttpMethod.Get, url);
+        request.Headers.Add("x-admin-secret", AdminSecret);
+        var response = await _httpClient.SendAsync(request);
+
+        if (response.IsSuccessStatusCode)
+        {
+            return await GetResult<CanceledMatchesResponse>(response);
+        }
+
+        return null;
+    }
+
     public async Task<List<PlayerWarningDefinition>> GetPlayerWarningDefinitions(bool includeDisabled = false)
     {
         var url = $"{MatchmakingApiUrl}/admin/warning-definitions";
@@ -761,6 +787,22 @@ public class PlayerWarningsGetRequest
     public int ItemsPerPage { get; set; } = 25;
     public string BattleTag { get; set; }
     public string Status { get; set; }
+}
+
+public class CanceledMatchesGetRequest
+{
+    public int Page { get; set; } = 1;
+    public int ItemsPerPage { get; set; } = 25;
+
+    // GameMode.Undefined (0) means "all game modes" and is not sent upstream.
+    public GameMode GameMode { get; set; }
+    public string BattleTag { get; set; }
+}
+
+public class CanceledMatchesResponse
+{
+    public int total { get; set; }
+    public List<Match> matches { get; set; }
 }
 
 public class PlayerWarningsResponse
--- a/W3C.Domain/MatchmakingService/MatchmakingServiceClient.cs
+++ b/W3C.Domain/MatchmakingService/MatchmakingServiceClient.cs
@@ -1,5 +1,6 @@
 ﻿using Newtonsoft.Json;
 using Newtonsoft.Json.Serialization;
+using Serilog;
 using System;
 using System.Collections.Generic;
 using System.Linq;
@@ -137,6 +138,9 @@ public class MatchmakingServiceClient
             return await GetResult<CanceledMatchesResponse>(response);
         }
 
+        var errorContent = await response.Content.ReadAsStringAsync();
+        Log.Error("Matchmaking service returned {StatusCode} fetching canceled matches (page {Page}, itemsPerPage {ItemsPerPage}, gameMode {GameMode}): {Content}",
+            response.StatusCode, req.Page, req.ItemsPerPage, req.GameMode, errorContent);
         return null;
     }
 
--- a/W3C.Domain/MatchmakingService/MatchmakingServiceClient.cs
+++ b/W3C.Domain/MatchmakingService/MatchmakingServiceClient.cs
@@ -117,11 +117,16 @@ public class MatchmakingServiceClient
 
     public async Task<CanceledMatchesResponse> GetCanceledMatches(CanceledMatchesGetRequest req)
     {
-        var url = $"{MatchmakingApiUrl}/admin/canceled-matches?page={req.Page}&itemsPerPage={req.ItemsPerPage}";
+        var url = $"{MatchmakingApiUrl}/admin/canceled-matches?itemsPerPage={req.ItemsPerPage}";
 
-        if (req.GameMode != GameMode.Undefined)
+        if (req.Cursor != null)
         {
-            url += $"&gameMode={(int)req.GameMode}";
+            url += $"&cursor={HttpUtility.UrlEncode(req.Cursor)}";
+        }
+
+        if (req.GameMode.HasValue)
+        {
+            url += $"&gameMode={(int)req.GameMode.Value}";
         }
 
         if (!string.IsNullOrEmpty(req.BattleTag))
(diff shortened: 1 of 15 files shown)


W3C.Domain/MatchmakingService/MatchmakingServiceClient.cs
```diff
@@ -763,6 +793,22 @@ public class PlayerWarningsGetRequest
     public string Status { get; set; }
 }
 
+public class CanceledMatchesGetRequest
+{
+    public int Page { get; set; } = 1;
+    public int ItemsPerPage { get; set; } = 25;
+
+    // GameMode.Undefined (0) means "all game modes" and is not sent upstream.
+    public GameMode GameMode { get; set; }
```
marcoabreu: Nullable then?

W3C.Domain/ReplayService/ReplayServiceClient.cs
```diff
@@ -15,9 +17,28 @@ public class ReplayServiceClient(IHttpClientFactory httpClientFactory)
...
+    // into DeepArchive (NotFound/Gone). Callers turn a null stream into a 404 rather
+    // than a 500. Any other non-success status is a genuine upstream failure (outage,
+    // timeout, etc.) and must not be mistaken for "no replay" -- it is logged and thrown
+    // so it surfaces as a server error instead of a misleading 404.
     public async Task<Stream> GenerateReplay(int gameId)
     {
-        var stream = await _httpClient.GetStreamAsync($"{ReplayServiceUrl}/generate/{gameId}?secret={AdminSecret}");
+        using var response = await _httpClient.GetAsync($"{ReplayServiceUrl}/generate/{gameId}?secret={AdminSecret}", HttpCompletionOption.ResponseHeadersRead);
+
+        if (response.StatusCode is HttpStatusCode.NotFound or HttpStatusCode.Gone)
```
marcoabreu, line 34: Could you please check which status code the replay service returns when the replay is archived?

W3ChampionsStatisticService/Admin/AdminMatchesController.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            Page = page,
+            ItemsPerPage = itemsPerPage,
+            GameMode = gameMode,
+            BattleTag = playerBattleTag,
+        });
+
+        if (result == null)
+        {
+            return StatusCode(502, "The matchmaking service did not return cancelled matches.");
```
marcoabreu, line 65: 404?

marcoabreu (approved):

W3C.Domain/MatchmakingService/MatchmakingServiceClient.cs
francislavoie: Done in `acf9ecd` — `GameMode?` on the request DTO and on the controller's query parameter, so "every mode" is its own value rather than enum member 0 doing double duty. Omitting the parameter now means omitting it, and the URL builder keys off `HasValue` instead of comparing against `Undefined`.

W3C.Domain/ReplayService/ReplayServiceClient.cs
francislavoie: Checked, and the answer was that it doesn't return one — the branch I'd written here could never have fired.

The replay service had a single `IntoResponse` that set **500 for every variant** of its error enum, including `GetGameArchiveFailed`, which is what an absent or archived replay produces. So "no replay" and "the service is broken" were indistinguishable to us, and a moderator opening an old match got a server error rather than a clean "no replay available".

Fixed at the source rather than worked around here: **w3champions/replay-service#12**. S3 already distinguishes the cases, so the fetch error is classified instead of flattened:

| S3 | Meaning | Status |
| --- | --- | --- |
| `NoSuchKey` | never written | 404 |
| `InvalidObjectState` | *"object is archived and inaccessible until restored"* — DEEP_ARCHIVE | 410 |
| anything else | credentials, dispatch, unparseable response | 500 |

Its game lookup had the same problem: a successful gRPC call returning no game (genuinely absent → 404) produced the same error as the call itself failing (→ 500). Split too.

The risk in that change is over-reaching the other way, so the mapping is allow-list: only the two S3 variants that positively mean absence become 4xx, everything unrecognised stays 500. An outage must never reach us as "no replay", or we'd show moderators a confident 404 for a game whose replay is fine. There's a test for that case specifically.

This side needed no logic change — both statuses were already handled — only the comment, which had recorded my assumption rather than the verified behaviour (`0069837`).

**Ordering:** replay-service#12 has to deploy before a 404 here can be trusted to mean "no replay". Until then the behaviour is what it is today.

W3ChampionsStatisticService/Admin/AdminMatchesController.cs
francislavoie: Kept as 502, but the question turned up a real bug next to it.

Why not 404: the request was well-formed and the collection exists — we just couldn't reach the only service that knows what's in it. A 404 says "there are no cancelled matches", which is a different claim, and one a moderator might act on: an empty moderation queue during a matchmaking outage looks like nothing to do. The reasoning is now a comment rather than tribal knowledge.

What it did surface: a genuinely bad request was also being flattened into 502. Matchmaking validates the game mode and (since matchmaking-service#866) the pagination cursor, so its 400 is the *caller's* error — a stale bookmark, a tampered cursor, an unknown mode. Blaming the upstream for that is wrong. A 400 is now rethrown and `HttpRequestExceptionFilter` turns it back into a 400 response; everything else still becomes 502 (`acf9ecd`).

Both paths have tests — `AnUnreachableMatchmakingServiceIsABadGatewayRatherThanANotFound` and `ARejectedCursorSurfacesAsABadRequestNotABadGateway`.

PR #556, opened 2026-08-12
Add CreatedAt and SeenAt to friend requests

MarkWind85:
## What

- `FriendRequest` gains nullable `CreatedAt` and `SeenAt`.
- `CreatedAt` is stamped server-side in `FriendCommandHandler.CreateFriendRequest` — `FriendRequest` is also a SignalR argument type, so a client-supplied value is overwritten.
- New hub method `MarkIncomingFriendRequestsSeen` (zero-arg, follows the existing `PreventZeroArgumentHandler` pattern): stamps `SeenAt` on every not-yet-seen request received by the caller, updates the in-memory cache coherently, and returns the refreshed received list to the caller. `SeenAt` records the *first* acknowledgement and never moves.

## Why

The launcher is growing a notification center, and friend requests are the one event where the client genuinely cannot infer what the player has already seen: a login with three pending requests looks identical whether they are new or week-old. With these fields the launcher can surface offline-received requests as unread and keep quiet about acknowledged ones.

## Compatibility

- Both fields nullable: existing Mongo documents deserialize with `null` ("predates the field"), no migration.
- Existing clients are unaffected — the new fields serialize alongside the existing ones and the new hub method is additive.
- Only unseen requests are touched by the mark-seen update (`SeenAt == null` filter), so repeated calls are idempotent.

## Tests

`MarkIncomingFriendRequestsSeen_StampsOnlyUnseenReceivedRequests` covers: unseen stamped, already-seen untouched, other receivers untouched, caller receives the refreshed list. All 39 hub tests pass.

### DIFF
--- a/W3ChampionsStatisticService/Friends/FriendCommandHandler.cs
+++ b/W3ChampionsStatisticService/Friends/FriendCommandHandler.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Linq;
 using System.Threading.Tasks;
 using MongoDB.Driver;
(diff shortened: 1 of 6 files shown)


MarkWind85:
Withdrawn — opened prematurely, will resubmit after internal review.

PR #557, opened 2026-08-12
Push a discriminated FriendChangeEvent for live friend events

MarkWind85:
## What

A new additive SignalR message, `FriendChangeEvent`, pushed to the affected player next to the existing `FriendResponseData` state push:

| Hub method | Pushed to | ChangeType | Actor |
|---|---|---|---|
| `MakeFriendRequest` | Receiver | `RequestReceived` | sender |
| `DeleteOutgoingFriendRequest` | Receiver | `RequestRetracted` | sender |
| `AcceptIncomingFriendRequest` | Sender | `RequestAccepted` | accepter |

`DenyIncomingFriendRequest` and `BlockIncomingFriendRequest` deliberately send **no** change event: they push the same `FriendResponseData` shape to the sender today and must stay indistinguishable (and "your request was denied" is not an event worth announcing — the request just disappears from the sent list).

## Why

`FriendResponseData` carries several distinct events under one message name, separated only by which arguments each call site leaves null. A client that wants to react to a *live* event (the launcher's notification center wiring "X sent you a friend request") can only infer it from that null-shape — an observation, not a contract, that a server refactor can change without any client failing to compile. `FriendChangeEvent` makes the event explicit and names the subject.

## Compatibility

Strictly additive: `FriendResponseData` is byte-for-byte unchanged, and clients that do not subscribe to `FriendChangeEvent` never see it.

## Tests

- `MakeFriendRequest_PushesFriendChangeEventToConnectedReceiver` — receiver gets `{RequestReceived, Sender#1}`.
- `DenyIncomingFriendRequest_SendsNoFriendChangeEventToSender` — the deliberate omission is pinned by a test.

All 40 hub tests pass.

### DIFF
--- a/W3ChampionsStatisticService/Friends/FriendChange.cs
+++ b/W3ChampionsStatisticService/Friends/FriendChange.cs
@@ -1,0 +1,24 @@
+namespace W3ChampionsStatisticService.Friends;
+
+// Wire values for FriendChange.ChangeType. String constants rather than an
+// enum so the serialized value is stable and self-describing for every client.
+public static class FriendChangeType
+{
+    public const string RequestReceived = "RequestReceived";
+    public const string RequestRetracted = "RequestRetracted";
+    public const string RequestAccepted = "RequestAccepted";
+}
+
+/// <summary>
+/// A discriminated live friend event, pushed as <c>FriendChangeEvent</c> to the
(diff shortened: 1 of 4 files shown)


MarkWind85:
Withdrawn — opened prematurely, will resubmit after internal review.

PR #563, opened 2026-09-17
fix: errors to run tests

godrien:
Errors were reported when running CI after the merge on main.
https://github.com/w3champions/website-backend/runs/104085102882

### Root Cause Analysis

1. **`KeyNotFoundException` in `PlayerGameLengthStats_GamesLengthsAreOK`:**
   - **Why it failed:** This test crashed because `GameLengthsByOpponentRace` lacked the `Race.UD` key. 
   - **Root Cause:** The `GameLengthsByOpponentRace` property was decorated with `[JsonIgnore]`. Due to recent changes in MongoDB Driver 3.x, `System.Text.Json.Serialization.JsonIgnore` is now honored natively by BSON serialization. As a result, the entire dictionary was ignored when being saved to the database. When a subsequent event loaded the document, the driver auto-initialized it to an empty dictionary, thereby erasing previously accumulated data for that player.
   - **Fix:** I explicitly appended `[BsonElement]` beneath the `[JsonIgnore]` attribute in `PlayerGameLength.cs` to force the MongoDB driver to serialize this data to the database while keeping it hidden from API JSON responses.

2. **`ValidationException: Clan not found` & `NullReferenceException` in Clan Tests (`PromoteShamanThatISNotInClan_Fails`, `CreateClan`, etc.):**
   - **Why it failed:** Tests that required `LoadClan` returned `null` for the Clan or loaded it with a completely null `ClanState`.
   - **Root Cause:** Similarly, `ClanState` inside `Clan.cs` had `[JsonIgnore]`. The MongoDB driver couldn't serialize/deserialize `ClanState` properly. Moreover, since `IntegrationTestBase.cs` didn't execute `BsonExtensions.AddSpecialBsonRegistrations`, the MongoDB driver was entirely unaware of the derived discriminator classes (`FoundedClan` and `NotFoundedClan`).
   - **Fix:** Added `[BsonElement]` to `ClanState` in `Clan.cs` and injected a static constructor into `IntegrationTestBase.cs` to invoke the `BsonExtensions.AddSpecialBsonRegistrations(services)` method, ensuring that BSON accurately registers these polymorphic class mappings during test suite execution.

3. **Flaky UI Test: `AdminJobRepositoryTests.DurationCountsWorkingTimeOnlyAcrossAPauseAndResume`:**
   - **Why it failed:** The test measures MongoDB operation intervals using very tight time bounds (adding exactly 250ms). Overheaded execution time inside the `Task.Delay` and consecutive `FinishAs` MongoDB calls inside standard local or CI machines usually pushes this total past the extremely strict tolerance point.
   - **Fix:** I increased the acceptable test duration variation bound to avoid brittle race conditions.


francislavoie:
okay so it just looks like random flaky stuff, nothing to do with your changes? Cool I guess :sweat_smile:

marcoabreu (approved):

